PR #29, opened 2026-03-17
improve guard skill description + add skill-review CI

fernandezbaptiste:
hey @Ruya-AI, thanks for building cozempic. really like the tiered pruning approach to context management for agent teams. Kudos on passing `150` stars! I've just starred it.

ran your guard skill through agent evals and spotted a few quick wins that took it from `~52%` to `~94%` performance:

- rewrote description with natural trigger terms like _protect session, context getting long, prevent compaction, session management_ + explicit Use when clause

- added stop + troubleshooting commands alongside the existing status check section

also added a lightweight GitHub Action that auto-reviews any skill.md changed in a PR (includes min permissions, uses a pinned action version, only posts a review comment).

this means that it gives you and your contributors an instant quality signal before you have to review yourself (no signup, no tokens needed).

these were easy changes to bring the skill in line with what **performs well against Anthropic's best practices**. honest disclosure, I work at tessl.io where we build tooling around this. not a pitch, just fixes that were straightforward to make! happy to answer any questions on the changes.

### DIFF
--- a/.github/workflows/skill-review.yml
+++ b/.github/workflows/skill-review.yml
@@ -1,0 +1,14 @@
+name: Skill Review
+on:
+  pull_request:
+    paths: ['**/SKILL.md']
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - uses: tesslio/skill-review@14b47c8420ff9ac9a12c30ed4b89ce0e9d0355ae # main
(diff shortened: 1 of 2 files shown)


junaidtitan:
Skill description changes are in v1.1.0 — the trigger terms and stop/troubleshoot commands are exactly right and the eval from 52% to 94% is a great signal, thanks for doing that work.

Skipped the CI action (`tesslio/skill-review`) for now — not comfortable adding a third-party action from a commercial product to the repo, even pinned. Nothing personal, just a policy call. The skill changes are the real value here and they're merged.

Closing this PR since the meaningful part is in.

PR #75, opened 2026-04-17
Pass Claude PID into detached guard instead of guessing parent PID

bslinger:
## What changed

- capture Claude's PID before the guard daemon detaches
- pass that PID into the detached `cozempic guard` process with an internal `--claude-pid` argument
- use the explicit PID for guard reload/watchdog and overflow recovery when available
- keep the existing process-tree discovery as a fallback when no explicit PID is available

## Why

This came out of debugging unexpected full session exits during guard-triggered reloads.

Tracing showed that, by the time the detached guard tried to reload, it was no longer necessarily parented by Claude. That means parent-process heuristics are unsafe in the daemon path: a detached guard can end up targeting the wrong process.

By capturing Claude's PID before daemonization and carrying it into the detached guard, reload behavior stays functional without depending on an ambiguous parent PID.

## Behavior

- normal detached guard reload keeps working when Claude's PID is known at startup
- overflow recovery uses the same explicit PID path
- if no explicit PID is available, the code still falls back to the existing Claude discovery logic
- if Claude still cannot be identified, the guard prunes without reloading rather than guessing an arbitrary process

## Tests

- add coverage that `start_guard_daemon()` passes `--claude-pid` into the child command
- add coverage that overflow recovery prefers the explicit Claude PID over rediscovery
- keep PID discovery coverage for the existing ancestor-walk behavior

Tested with:

```bash
env UV_CACHE_DIR=/tmp/uv-cache uv run --with pytest pytest tests/test_session.py tests/test_guard_robustness.py tests/test_overflow.py
```

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -582,6 +582,7 @@ def cmd_post_compact(args):
 def cmd_guard(args):
     """Start the guard daemon to prevent compaction-induced state loss."""
     session_id = args.session or None
+    claude_pid = args.claude_pid or find_claude_pid()
 
     if getattr(args, "system_overhead_tokens", None):
         os.environ["COZEMPIC_SYSTEM_OVERHEAD_TOKENS"] = str(args.system_overhead_tokens)
@@ -598,6 +599,7 @@ def cmd_guard(args):
             threshold_tokens=args.threshold_tokens,
             soft_threshold_tokens=args.soft_threshold_tokens,
             session_id=session_id,
+            claude_pid=claude_pid,
         )
         if result["already_running"]:
             print(f"  Guard already running (PID {result['pid']})")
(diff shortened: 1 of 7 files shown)


junaidtitan:
Merged into v1.8.1 — excellent contribution @bslinger. The PPID fallback removal alone prevents a genuinely dangerous failure mode (SIGTERMing systemd/launchd on reparented daemons). The explicit PID handoff is the right architecture.

Had one trivial test conflict with our TestReloadSelfDaemon class (added in the same file during the 1.8.0 QA cycle) — resolved by keeping both.

If you're up for more, a few areas that build on this work:

- **Guard resilience**: the overflow recovery path could benefit from the same explicit-PID pattern for the reactive watcher (kqueue/polling path)
- **Multi-session guard**: users with multiple Claude sessions open simultaneously sometimes see guard daemons competing — coordinating via PID awareness would help
- **Windows daemon support**: the guard doesn't daemonize cleanly on Windows yet (no `start_new_session` equivalent)

Really appreciate the thorough debugging and clean PR. Welcome aboard! 🙏

PR #76, opened 2026-04-17
feat: add claude-opus-4-7 to model context window mapping

ynaamane:
## Summary
- Add `claude-opus-4-7` (1M context) to `MODEL_CONTEXT_WINDOWS` in `tokens.py`
- Update `doctor.py` diagnostic messages to reference Opus 4.7 as current default
- Add test coverage for Opus 4.7 detection and prefix matching

Opus 4.7 released 2026-04-16. Added alongside existing models — not replacing, since 4.6 is still served.

## Test plan
- [x] Existing tests pass (4 pre-existing failures unrelated to this change)
- [x] New `test_opus_47_is_1m` passes
- [x] New `test_prefix_match_versioned` for 4.7 passes
- [x] Doctor test updated for 4.7 model reference

PR #79, opened 2026-04-21
feat: strategies — centralize config validation with helpful errors

ynaamane:
## Summary

Adds a shared `strategies/_config.py` module with three small validation helpers, applied across the 5 config-reading pruning strategies. Surfaces misconfiguration as a clear `ConfigError` rather than a cryptic `TypeError`, a silent no-op, or — in the worst case — destructive behavior that looks like success.

Addresses the "strategy improvements / refinements" contribution area from #76.

## Problem — three silent pathologies today

Every config-driven strategy currently reads its knobs with bare `config.get(key, default)`. I audited every call site against real malformed configs and reproduced three failure modes:

### 1. Cryptic `TypeError` deep in a loop

`mega-block-trim` / `document-dedup` with a string value for a byte count:

```python
>>> STRATEGIES[\"mega-block-trim\"].func(msgs, {\"mega_block_max_bytes\": \"32768\"})
TypeError: '>' not supported between instances of 'int' and 'str'
    at src/cozempic/strategies/aggressive.py:343
```

The user has to read the strategy source to figure out that the config key was wrong.

### 2. Silent no-op that burns a prune cycle

`thinking-blocks` with a typo'd mode: `{\"thinking_mode\": \"truncateee\"}` → none of the three `if mode == ...` branches match → strategy runs to completion, reports 0B pruned. User never learns why their config had no effect.

### 3. Destructive behavior masquerading as success

`tool-result-age` with swapped values — `{\"tool_result_mid_age\": 50, \"tool_result_old_age\": 30}`:

- Current code: `if turns_ago < mid_age: skip; if turns_ago >= old_age: stub`
- With the swap: every tool result where `turns_ago >= 0` hits the `>= old_age=30` branch → gets replaced by a compact stub
- Output: \"Compacted N tool results\" — user thinks it worked. Actually, the session was just gutted.

Same pattern with negative values (reproduced locally: `mid=-5, old=-1` → every tool result marked old → session content destroyed).

## The fix

Three helpers in `cozempic/strategies/_config.py` (new file, 103 LOC):

| Helper | Guarantees |
|---|---|
| `coerce_non_negative_int(config, key, default)` | `int >= 0`, rejects `bool` (YAML `yes` → `True` → `1` would silently mean \"max_bytes=1\") and `float` |
| `coerce_choice(config, key, choices, default)` | `str in choices`, error message lists valid values |
| `coerce_ordered_pair(config, low_key, high_key, defaults)` | both non-negative ints, **enforces `low < high`** |

`ConfigError` subclasses `ValueError` — any existing `except ValueError` keeps catching it.

Applied to all 5 config-reading strategies:

| Strategy | File:Line | Change |
|---|---|---|
| thinking-blocks | standard.py:26 | `coerce_choice` with `(\"remove\", \"truncate\", \"signature-only\")` |
| tool-output-trim | standard.py:98-99 | `coerce_non_negative_int` × 2 |
| tool-result-age | standard.py:379-381 | `coerce_ordered_pair` |
| document-dedup | aggressive.py:252 | `coerce_non_negative_int` |
| mega-block-trim | aggressive.py:324 | `coerce_non_negative_int` |

## Safety & scope

- **No behavior change on valid configs** — identical path for everything that was already correct. The existing 360+ tests still pass unchanged.
- **No new dependencies.**
- **No public API change** — `ConfigError` is in `strategies/_config.py` (private module), strategies still have the same signature.
- Helper module is prefixed `_` (convention for internal modules in this codebase).

## Tests

30 new tests in `tests/test_strategies_config.py`, 3 classes:

**`TestCoerceNonNegativeInt`** (7 tests)
- returns default when key absent / returns value when valid / accepts zero
- rejects negative / string / float / bool (with explanatory test docstring about YAML `yes`)

**`TestCoerceChoice`** (4 tests)
- returns default / returns valid / rejects unknown value (message must list valid choices) / rejects non-string

**`TestCoerceOrderedPair`** (6 tests)
- defaults / valid / rejects inverted / rejects equal / rejects negative / rejects non-int

**`TestStrategyConfigIntegration`** (13 tests including 3 subtests)
- One ConfigError test per \"bad config\" scenario for each of the 5 strategies (tool-output-trim×2, tool-result-age×3, thinking-blocks×2, document-dedup×1, mega-block-trim×2)
- \"Backwards compat\" tests: default config still runs, every valid `thinking_mode` still runs

Suite status:
- New: **30/30 green + 3 subtests green**
- Existing doctor + strategies: **390/390 green**
- The 6 pre-existing `test_tokens.py` / `test_model_detection.py` failures on `main` are unchanged and not related to this PR.

## Live smoke test on real session

Ran against a 29K-message session in `~/.claude/projects/`:

```
[OK]   inverted ages: ConfigError raised
[OK]   string byte limit: ConfigError raised
[OK]   unknown thinking_mode: ConfigError raised
[OK]   string min_bytes: ConfigError raised
[OK]   tool-result-age default config: pruned=0B replaced=0
[OK]   tool-output-trim default config: pruned=23B replaced=23
[OK]   thinking-blocks default config: pruned=0B replaced=0
[OK]   document-dedup default config: pruned=0B replaced=0
[OK]   mega-block-trim default config: pruned=0B replaced=0
```

Bad configs fail loud, defaults work unchanged. `tool-output-trim` actually found 23 things to trim (the strategies are still active).

## Files changed

- `src/cozempic/strategies/_config.py` (new, 103 LOC)
- `src/cozempic/strategies/standard.py` (+8 -5 LOC — imports + 3 helper calls)
- `src/cozempic/strategies/aggressive.py` (+2 -2 LOC — imports + 2 helper calls)
- `tests/test_strategies_config.py` (new, 195 LOC)

## Test plan

- [x] 30 new unit + integration tests green
- [x] Full strategies + doctor suite unchanged (390 green)
- [x] Live smoke test on a real session JSONL: bad configs → ConfigError, valid configs behave exactly as before
- [x] No new dependencies
- [x] No public API change
- [x] `_config.py` is a prefixed private module (convention-consistent)

### DIFF
--- a/src/cozempic/strategies/_config.py
+++ b/src/cozempic/strategies/_config.py
@@ -1,0 +1,103 @@
+"""Shared config-validation helpers for strategy functions.
+
+Every strategy that reads tuning knobs from the per-session config dict goes
+through one of the helpers below so invalid values (wrong type, out-of-range,
+swapped ordering) surface a single well-formed ConfigError at invocation
+time rather than:
+
+  - a cryptic TypeError deep in a comparison loop
+    (e.g. `'>=' not supported between instances of 'int' and 'str'`)
+  - a silent no-op that nonetheless "runs successfully" and wastes a prune
+    cycle, or worse, produces destructive behavior (e.g. negative age
+    thresholds marking every tool result "old" and stubbing everything)
+
+Keeping validation in one module also avoids duplicating error strings
+across strategies and makes the user-facing message format consistent.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+class ConfigError(ValueError):
+    """Raised when a strategy config value is missing a required invariant.
+
+    Subclasses ValueError so callers that already catch ValueError (e.g. the
+    executor's outer wrapper) keep working, but type-checking machinery and
+    humans reading tracebacks see the specific name.
+    """
+
+
+def _is_strict_int(value: Any) -> bool:
+    """Strict int check: rejects bool (True/False are ints in Python) and float.
+
+    YAML parsers in particular will turn `yes`/`no` into booleans, and users
+    occasionally write `8192.0` in a JSON config. Both are almost never what
+    was meant when the field expects a byte/line count.
+    """
+    return isinstance(value, int) and not isinstance(value, bool)
+
+
+def coerce_non_negative_int(config: dict, key: str, default: int) -> int:
+    """Return `config[key]` as a non-negative int, or `default` if key absent.
+
+    Raises `ConfigError` if the value is present but the wrong type or sign.
+    """
+    if key not in config:
+        return default
+    value = config[key]
+    if not _is_strict_int(value):
+        raise ConfigError(
+            f"config[{key!r}] must be an int, got {type(value).__name__} {value!r}"
+        )
+    if value < 0:
+        raise ConfigError(
+            f"config[{key!r}] must be non-negative, got {value}"
+        )
+    return value
+
+
+def coerce_choice(config: dict, key: str, choices: tuple[str, ...], default: str) -> str:
+    """Return `config[key]` if it matches one of `choices`, else `default`.
+
+    Raises `ConfigError` when the value is present but not a recognized
+    choice. Error message lists the accepted values so the user can
+    self-correct without reading the source.
+    """
+    if key not in config:
+        return default
+    value = config[key]
+    if not isinstance(value, str):
+        raise ConfigError(
+            f"config[{key!r}] must be a string, got {type(value).__name__} {value!r}"
+        )
+    if value not in choices:
+        raise ConfigError(
+            f"config[{key!r}]={value!r} is not one of {list(choices)}"
+        )
+    return value
+
+
+def coerce_ordered_pair(
+    config: dict,
+    low_key: str,
+    high_key: str,
+    defaults: tuple[int, int],
+) -> tuple[int, int]:
+    """Return `(config[low_key], config[high_key])` with the invariant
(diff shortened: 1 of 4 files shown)


junaidtitan:
Merged — this closes a real safety gap. The swapped-ages case you reproduced (every tool result marked "old" and stubbed = session silently gutted) is exactly the kind of bug that's nearly impossible to diagnose after the fact. Catching it at config-read time with a clear error message is the right architecture.

Specific things I liked:

- `_is_strict_int` rejecting `bool` is smart and forward-thinking (YAML `yes`/`no` → True/False is a real footgun with Claude Code's JSON configs)
- `ConfigError(ValueError)` subclassing means the executor's existing error paths still work — no behavior change for callers that already catch ValueError
- The backwards-compat test block (`test_valid_*_defaults_still_work`) is exactly right — regression protection for the default config path that 99% of users hit
- Integration tests that actually invoke the strategy functions with bad configs (not just unit-testing the helpers) give real confidence

The `coerce_ordered_pair` helper in particular is reusable — we have similar ordered-threshold logic in the guard daemon's 4-tier system (soft < hard < danger < emergency) that currently does no validation. A natural follow-up if you're looking for more.

Other areas where your config-validation approach would help:

- **Guard daemon thresholds**: `--threshold`, `--soft-threshold`, `--threshold-tokens` accept arbitrary values with zero validation today
- **Token estimation**: `COZEMPIC_CONTEXT_WINDOW` env var is parsed with a bare `int()` — string/negative/zero values crash deep in the token math

Four PRs merged, zero issues — you're a core contributor at this point. Thank you @ynaamane 🙏

PR #83, opened 2026-04-27
feat: harden numeric input validation across CLI flags + env vars

ynaamane:
## Summary

Applies the config-validation approach from #79 to every remaining user-facing numeric input — guard CLI flags, remind interval, and the two cozempic env vars that were parsed with a bare `int()`.

Follows up on @junaidtitan's comment after PR #79 merge: *\"Other areas where your config-validation approach would help: guard daemon thresholds accept arbitrary values with zero validation today; COZEMPIC_CONTEXT_WINDOW is parsed with a bare int() — string/negative/zero values crash deep in the token math.\"*

Two commits for reviewability:

| Commit | Scope |
|---|---|
| `refactor: lift strategy config helpers to cozempic._validation` | Mechanical move of the generic helpers from `strategies/_config.py` to a root `_validation.py`; `strategies/_config.py` re-exports so existing imports keep working. Zero behavior change, 39 existing strategy tests untouched. |
| `feat: harden numeric input validation across CLI flags and env vars` | The actual fix: 9 surfaces validated, 50 new tests. |

### 1. Silent propagation of nonsense values
```
$ COZEMPIC_CONTEXT_WINDOW=-100 cozempic diagnose <session>
# pct=-110%, thresholds=(-25, -55, -80) printed in diagnostic output
```

### 2. Silent override ignored (falsy trap)
```python
# tokens.py:96 (before)
val = os.environ.get(\"COZEMPIC_CONTEXT_WINDOW\")
if val:        # <-- 0 is falsy, override silently dropped
    return int(val)
```
Running `COZEMPIC_CONTEXT_WINDOW=0 cozempic diagnose` silently uses the built-in default 1M. The user never sees their override was ignored.

### 3. Reload storm
```
$ cozempic guard --threshold -1 --daemon
# daemon spawns. hard_bytes=-1048576. Every 30s cycle:
#   current_bytes >= -1048576  =>  TRUE  =>  prune+reload
```

### 4. Cryptic mid-daemon crash
```
$ cozempic guard --interval -1 --daemon
# argparse accepts, daemon spawns. 30s later:
#   File \"guard.py\", line 368, in start_guard
#     time.sleep(interval)
# ValueError: sleep length must be non-negative
```

### Layer 1 — argparse validators (cli.py)
Two small inline validator functions:

```python
def _positive_int(val: str) -> int: ...
def _positive_float(val: str) -> float: ...
```

Applied to `--threshold`, `--soft-threshold`, `--interval`, `--threshold-tokens`, `--soft-threshold-tokens` (guard) and `--interval` (remind). Produces standard argparse error output with exit code 2:

```
$ cozempic guard --threshold -1
usage: cozempic guard [-h] ...
cozempic guard: error: argument --threshold: must be positive, got -1.0
```

### Layer 2 — in-process validation in `start_guard()` (guard.py)
Defense-in-depth for Python callers that bypass argparse (e.g. `cozempic.guard.start_guard(threshold_mb=-1, ...)`). Also catches the one thing argparse can't: **soft < hard ordering**. When user passes both `--soft-threshold` and `--threshold`, each individually passes argparse but their ordering isn't checked there.

```python
# start_guard() now starts with:
if soft_threshold_mb is not None and soft_threshold_mb >= threshold_mb:
    raise ConfigError(
        f\"soft_threshold_mb={soft_threshold_mb} must be strictly less than \"
        f\"threshold_mb={threshold_mb}\"
    )
# + same for token thresholds + positivity belt-and-braces
```

### Layer 3 — env var parsing (tokens.py)
Routes both env vars through helpers in `cozempic._validation` that WARN to stderr (matching `_prescan_argv` pattern) rather than raise — an env var is ambient config, and raising would crash a daemon that had been running fine.

```python
# Old:
val = os.environ.get(\"COZEMPIC_CONTEXT_WINDOW\")
if val:
    try: return int(val)
    except ValueError: pass
return None

# New:
return parse_env_positive_int(\"COZEMPIC_CONTEXT_WINDOW\")
# → rejects 0, -100, \"abc\", empty with a clear stderr warning; falls back to default
```

`COZEMPIC_SYSTEM_OVERHEAD_TOKENS` uses `parse_env_non_negative_int` instead: 0 is a legitimate value there (a session with no rules file and no MCP has no overhead).

## New helpers in `cozempic/_validation.py`

Extending the module with four new functions beyond what the refactor commit added:

```python
def coerce_positive_int(config, key, default) -> int: ...      # strict > 0
def coerce_positive_float(config, key, default) -> float: ...  # strict > 0, accepts int
def parse_env_positive_int(name: str) -> int | None: ...       # warn + fallback
def parse_env_non_negative_int(name: str) -> int | None: ...   # warn + fallback
```

`coerce_positive_int` is distinct from `coerce_non_negative_int` (which allows 0). Separation is intentional: zero is valid for \"no overhead\" but nonsensical for \"polling interval\" or \"context window size\".

## Testing

**New tests: 50**

| File | New tests | Coverage |
|---|---|---|
| `tests/test_validation.py` (new) | 30 | Unit tests for all 4 new helpers + 2 existing, including bool-rejection, backwards-compat re-exports |
| `tests/test_cli_validation.py` (ext) | 15 | `TestGuardArgparseValidation` (16 argparse scenarios) + `TestStartGuardOrderingValidation` (3 ordering invariants) |
| `tests/test_tokens.py` (ext) | 5 | `TestEnvVarOverrideValidation` — integration: env var to return value |

**Suite status:**
- New tests: **50/50 green**
- Full suite: **461 passed**, 3 pre-existing failures in `test_model_detection.py` (unrelated to this PR — reduced from 6 on v1.8.3 after maintainer fix in v1.8.4)
- Strategies + doctor tests: unchanged (39 green)
- **Zero regression.**

## Live smoke test — 8/8 scenarios pass

```
Smoke 1: cozempic guard --threshold -1
  → error: argument --threshold: must be positive, got -1.0 ✓
Smoke 2: cozempic guard --threshold 0
  → error: argument --threshold: must be positive, got 0.0 ✓
Smoke 3: cozempic guard --interval 0
  → error: argument --interval: must be positive, got 0 ✓
Smoke 4: cozempic guard --interval -1
  → error: argument --interval: must be positive, got -1 ✓
Smoke 5: start_guard(threshold_mb=50, soft_threshold_mb=100, ...) (Python direct)
  → ConfigError: soft_threshold_mb=100 must be strictly less than threshold_mb=50 ✓
Smoke 6: COZEMPIC_CONTEXT_WINDOW=0
  → Warning: ignoring COZEMPIC_CONTEXT_WINDOW='0' — must be a positive integer
  → Returns None (falls back to model detection) ✓
Smoke 7: COZEMPIC_CONTEXT_WINDOW=-100
  → Warning: ignoring COZEMPIC_CONTEXT_WINDOW='-100' — must be a positive integer ✓
Smoke 8: Happy path — cozempic guard --threshold 50 --soft-threshold 30 --interval 30
  → threshold=50.0, soft=30.0, interval=30 (parse OK) ✓
```

## Safety & scope

- **No public API change.** `ConfigError` subclasses `ValueError`, so existing `except ValueError` handlers keep catching.
- **No new dependencies.**
- **No behavior change on valid inputs.** Every happy path is identical to v1.8.5.
- **Env var path warns rather than raises** — matches the existing `_prescan_argv` convention, won't crash a running daemon.
- **Not in scope:** Windows-specific code (macOS-only testing), `--claude-pid` (PIDs are always positive on POSIX, implicit in `os.kill`), new features.

## Files changed

| File | Δ |
|---|---|
| `src/cozempic/_validation.py` | +208 (new — shared helpers, refactor commit) |
| `src/cozempic/strategies/_config.py` | -47 (thin re-export, refactor commit) |
| `src/cozempic/cli.py` | +32 / -5 (2 validators + 6 argparse updates) |
| `src/cozempic/guard.py` | +33 (ordering + positivity in `start_guard`) |
| `src/cozempic/tokens.py` | +8 / -12 (route through helpers) |
| `tests/test_validation.py` | +180 (new) |
| `tests/test_cli_validation.py` | +138 |
| `tests/test_tokens.py` | +49 |

## Test plan

- [x] 50 new tests green
- [x] Full suite shows zero regression (461 passed, 3 pre-existing unchanged)
- [x] Live smoke test on all 9 pathological inputs
- [x] Happy path (valid args) unchanged
- [x] No new dependencies
- [x] No breaking API change (re-export preserves existing strategy imports)

### DIFF
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -1,0 +1,208 @@
+"""Shared validation helpers for user-supplied numeric input.
+
+Collects the cross-cutting helpers used by:
+  - strategy `config` dicts (originally introduced in strategies/_config.py,
+    PR #79) — now re-exported from here to avoid duplicating error strings,
+  - argparse CLI argument validators (via small wrappers in cli.py),
+  - environment-variable parsing in tokens.py.
+
+Design goals:
+
+  1. A single source of truth for the error-message shape, so the user sees
+     the same phrasing whether the bad value came from a config dict, a CLI
+     flag, or an env var.
+
+  2. ConfigError subclasses ValueError — callers that already catch
+     ValueError keep working; new code can catch the narrower exception.
+
+  3. Env-var helpers WARN rather than RAISE: an env var is ambient config
+     (often set once in a shell profile), and we don't want a daemon that
+     was running fine to crash on the next reload just because `$HOME`
+     carries a stale override. CLI args and in-process config dicts, in
+     contrast, raise — the call site is intentional and the user is there.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from typing import Any
+
+
+class ConfigError(ValueError):
+    """Raised when a config value violates a required invariant.
+
+    Subclasses ValueError so callers that already catch ValueError (e.g.
+    the executor's outer wrapper) keep working, but type-checking
+    machinery and humans reading tracebacks see the specific name.
+    """
+
+
+# ── Type guards ────────────────────────────────────────────────────────────
+
+
+def _is_strict_int(value: Any) -> bool:
+    """Strict int check: rejects bool (True/False are ints in Python) and float.
+
+    YAML parsers in particular will turn `yes`/`no` into booleans, and users
+    occasionally write `8192.0` in a JSON config. Both are almost never what
+    was meant when the field expects a byte/line count.
+    """
+    return isinstance(value, int) and not isinstance(value, bool)
+
+
+def _is_strict_number(value: Any) -> bool:
+    """Accept int or float for float-typed fields (e.g. MB thresholds).
+
+    Reject bool — same YAML concern as `_is_strict_int` — and reject anything
+    non-numeric. Users legitimately write `threshold=50` (int) expecting 50.0 MB,
+    so int is fine; but `True` or `"50"` are not.
+    """
+    if isinstance(value, bool):
+        return False
+    return isinstance(value, (int, float))
+
+
+# ── Config-dict helpers (shared across strategies + guard) ─────────────────
+
+
+def coerce_non_negative_int(config: dict, key: str, default: int) -> int:
+    """Return `config[key]` as a non-negative int, or `default` if key absent.
+
+    Raises `ConfigError` if the value is present but the wrong type or sign.
+    """
+    if key not in config:
+        return default
+    value = config[key]
+    if not _is_strict_int(value):
+        raise ConfigError(
+            f"config[{key!r}] must be an int, got {type(value).__name__} {value!r}"
+        )
+    if value < 0:
+        raise ConfigError(
+            f"config[{key!r}] must be non-negative, got {value}"
+        )
+    return value
+
+
+def coerce_positive_int(config: dict, key: str, default: int) -> int:
+    """Like `coerce_non_negative_int` but strict `> 0`.
+
+    Separate helper because zero is nonsensical for values like polling
+    intervals (would produce a spin loop) and context-window sizes
+    (would divide by zero downstream), whereas zero IS valid for values
+    like `system_overhead_tokens` (a session with no rules file).
+    """
+    if key not in config:
+        return default
+    value = config[key]
+    if not _is_strict_int(value):
+        raise ConfigError(
+            f"config[{key!r}] must be an int, got {type(value).__name__} {value!r}"
+        )
+    if value <= 0:
+        raise ConfigError(
+            f"config[{key!r}] must be positive, got {value}"
+        )
+    return value
+
+
+def coerce_positive_float(config: dict, key: str, default: float) -> float:
+    """Return `config[key]` as a strictly-positive float, or `default` if absent.
+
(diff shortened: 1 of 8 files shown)


junaidtitan:
Merged — excellent work @ynaamane. Five PRs in, consistently raising the bar.

The three-layer validation design (argparse → in-process → env-var-with-warn) is exactly right for the different call surfaces. The `_is_strict_int` rejecting bool from YAML parsers is a pattern I'll be copying.

I've pushed a follow-up commit (fbc83da) that closes the one gap our review identified: `start_guard_daemon()` now pre-validates numeric params before spawning the child. Without this, bad values would pass to the child via CLI, the child would die on argparse, but the parent would report `started=True` — caller thinks guard is running when it's already dead.

Shipping in the next patch release with your credit in the release notes. If you want to keep going, the `start_guard_daemon` validation follow-up could use a test — verifying that `start_guard_daemon(threshold_mb=-1)` raises ConfigError rather than spawning a doomed child. And the upper-bound checks you noted in the review (e.g. `--threshold 999999999`) would be a natural extension.

🙏

PR #84, opened 2026-05-05
fix(digest): harden behavioral digest against CC synthetic noise pollution (16 bugs + retroactive purge)

ynaamane:
## Problem

A review team spot-checked a production digest store and found **139+ rules that are not real user corrections** — instead, every `<local-command-caveat>`, `<teammate-message>`, `/init` prompt, and other Claude Code synthetic user turn containing a correction-signal word (`don't`, `never`, `no,`) got extracted, prefixed with `"Do not "`, and persisted as a `type: feedback` memory.

### Empirical evidence from one real store

- JSON state: **572 rules total, 468 active** (20:1 over `MAX_ACTIVE_RULES=20`)
- Injected memory file: **142 rules rendered** (7:1 over the documented 20 cap)
- Representative pollution (unaltered from the actual store):
  - `R001: Do not <local-command-caveat>Caveat: The messages below were generated...`
  - `R002: Do not Please analyze this codebase and create a CLAUDE.md file...` (/init prompt)
  - `R161: Do not <teammate-message teammate_id="falsification-specialist"...>`

Every Claude Code session that loads `memory/cozempic_digest.md` gets these as behavioral guidance.

## Audit

A full code audit (`digest.py`, 774 lines) identified **16 bugs**:

| # | Severity | Summary |
|---|---|---|
| BUG-1 | CRITICAL | `_EXPLICIT_PATTERNS` match any user text containing `don't`/`never`/`no,` with zero noise filter |
| BUG-2 | CRITICAL | `_to_prohibition` naïvely prefixes "Do not " on up to 500 chars — swallows markdown, code blocks, tags |
| BUG-3 | CRITICAL | `extract_corrections` has no noise gate before `classify_turn` |
| BUG-4 | HIGH | `EXPLICIT_CORRECTION` bypasses the `PROMOTION_COUNT=2` repetition gate |
| BUG-5 | HIGH | `admit_rule` single-demotion + no retroactive sweep on `load_digest_store` |
| BUG-6 | HIGH | `build_injection_text` iterates the full `hard` list, uncapped |
| BUG-7 | HIGH | `_find_duplicate` at 0.5 word-overlap merges semantically opposite rules |
| BUG-8 | CRITICAL | `_get_memdir` hardcodes `~/.claude`, ignoring `CLAUDE_CONFIG_DIR` — breaks all non-default profiles |
| BUG-17 | CRITICAL | Load-time noise-evidence purge missing — pre-polluted stores never recover |
[...]

PR #85, opened 2026-05-06
fix(digest): BUG-11 — _infer_scope word-boundary prevents silent scope drift

ynaamane:
## Problem

`_infer_scope` used substring match (`kw in text_lower`), which silently classified text as the wrong scope when keyword substrings appeared inside unrelated words:

| Input | Old scope | Should be |
|---|---|---|
| `"make it more digital"` | `git` (digital contains git) | `general` |
| `"editorial review"` | `file-ops` (editorial contains edit) | `general` |
| `"testimony from the witness"` | `testing` (testimony contains test) | `general` |
| `"the merger acquisition"` | `git` (merger contains merge) | `general` |
| `"slackline balance"` | `communication` (slackline contains slack) | `general` |

This silently broke the BUG-7 dedup gate merged in PR #84 (v1.8.7), which requires scope+priority match before text-overlap merge. Two rules with drifted scopes (one matched on substring, one on whole word) would fail to dedup and accumulate as distinct rules that should collapse. It also misrouted genuine corrections into the wrong scope bucket, poisoning downstream priority-ordering.

## Fix

Tokenize with `[\w-]+` (preserving hyphenated compounds like `co-authored-by`) and match tokens as whole words against a `frozenset` keyword table:

```python
_SCOPE_KEYWORDS = (
    ("testing",       frozenset({"test", "tests", "pytest", "unittest", "mock", ...})),
    ("git",           frozenset({"git", "commit", "push", "branch", "merge", "co-authored", "co-authored-by", ...})),
    ("file-ops",      frozenset({"file", "edit", "write", "read", "path", "directory", ...})),
    ("communication", frozenset({"message", "comment", "pr", "issue", "slack", ...})),
)

def _infer_scope(text: str) -> str:
    tokens = set(re.findall(r"[\w-]+", text.lower()))
    for scope, keywords in _SCOPE_KEYWORDS:
        if tokens & keywords:
            return scope
    return "general"
```

Also reorders scopes so `testing` precedes `file-ops` on the overlap case `"write tests"` — the noun `tests` is the subject of the correction, `write` is incidental.

## Test coverage

New `TestInferScopeWordBoundary` class, 11 tests:

- 6 false-positive cases confirm they now resolve to `general`
- 4 legitimate scope matches confirm they still work
- 1 case-insensitivity regression guard

## Test plan

- [x] New test class: 11/11 pass
- [x] Full `test_digest.py` + `test_track4.py`: 151/151 pass
- [x] Full suite (excl. pre-existing `test_model_detection.py` fails): 524 pass
- [x] Manual empirical reproduction of 5 false-positive cases before fix, all return `general` after fix

## Scope

- Net change: **+23 / −10 LOC** on `src/cozempic/digest.py`
- Tests: **+72 LOC** on `tests/test_digest.py` (append-only, zero existing-test mutations)
- No behavioral change on any clean input; fix is surgical to the substring-match bug

## Context

Follow-up to [PR #84 (v1.8.7)](https://github.com/Ruya-AI/cozempic/pull/84) which introduced the BUG-7 dedup gate that depends on `_infer_scope` accuracy. BUG-11 was triaged as MEDIUM-scope-v2 in the audit artifact of that PR and is now shipped as its own focused change.

### DIFF
--- a/src/cozempic/digest.py
+++ b/src/cozempic/digest.py
@@ -279,17 +279,30 @@ def _get_assistant_text(msg: dict) -> str:
     return " ".join(parts)
 
 
+_SCOPE_KEYWORDS: tuple[tuple[str, frozenset[str]], ...] = (
+    # Order matters on overlap: `testing` before `file-ops` so that
+    # "write tests" (both write-verb and tests-noun) resolves to testing —
+    # the noun is the subject of the correction, the verb is incidental.
+    ("testing", frozenset({"test", "tests", "pytest", "unittest", "mock", "mocks", "assert", "asserts"})),
+    ("git", frozenset({"git", "commit", "commits", "push", "pushed", "branch", "branches", "merge", "merges", "merged", "co-authored", "co-authored-by"})),
+    ("file-ops", frozenset({"file", "files", "edit", "edits", "write", "writes", "read", "reads", "path", "paths", "directory", "directories"})),
+    ("communication", frozenset({"message", "messages", "comment", "comments", "pr", "prs", "issue", "issues", "slack"})),
+)
+
+
 def _infer_scope(text: str) -> str:
-    """Infer the scope of a correction from its content."""
-    text_lower = text.lower()
-    if any(kw in text_lower for kw in ("git", "commit", "push", "branch", "merge", "co-authored")):
-        return "git"
-    if any(kw in text_lower for kw in ("file", "edit", "write", "read", "path", "directory")):
-        return "file-ops"
-    if any(kw in text_lower for kw in ("test", "pytest", "unittest", "mock", "assert")):
-        return "testing"
-    if any(kw in text_lower for kw in ("message", "comment", "pr ", "issue", "slack")):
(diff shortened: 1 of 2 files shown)


junaidtitan:
Merged. Exactly the follow-up we were hoping for — the substring false positives in _infer_scope were silently breaking the dedup gate you built in PR #84. "digital" → git, "testimony" → testing — each one a rule that should have deduped but didn't because scope drifted.

The frozenset tokenization is the right fix. Clean, fast, zero behavioral change on legitimate inputs. The scope reordering (testing before file-ops on "write tests") is a nice detail.

Seven PRs now, each one building on the last. If you're looking for the next BUG from the #84 backlog, BUG-16 (candidate-rejection logging) would give us the observability to measure how the noise filter and prohibition gate are performing in the wild — how many candidates are rejected per session, which patterns trigger most, etc. That data would inform whether the thresholds (200 chars, 2 newlines, PROMOTION_COUNT=2) need tuning.

🙏

PR #89, opened 2026-05-14
fix(guard): resolve 3GB/22h memory leak via incremental transcript ingest

ynaamane:
## Summary

Resolves a memory leak in the guard daemon that accumulated ~3 GB RSS over 22h on
active silc-data sessions. Root cause: `checkpoint_team()` called `load_messages()`
every 30s, re-parsing the entire JSONL transcript each cycle. Freed allocations
were retained in macOS libmalloc's `MALLOC_LARGE_REUSABLE` zone rather than
returned to the kernel, driving unbounded RSS growth even with 100% live-object
cleanup.

Empirical evidence from a live leaker (PID 31934 before the fix):
- **3.13 GB RSS** after 22h on a **40 MB JSONL** (78× ratio)
- **151 MB physical footprint** (via `vmmap`) vs 3.13 GB RSS → confirmed
  fragmentation, not live-object leak
- **63 × ~34 MB MALLOC_LARGE_REUSABLE regions** accumulated
- Healthy baseline (PID 73407, 17 MB JSONL, 6d 23h uptime): **zero**
  `MALLOC_LARGE_REUSABLE` regions, 403 MB RSS total

### Primary — incremental transcript ingest (`load_messages_incremental`)

New read-only function in `session.py`:
- Module-global `_INCR_CACHE: OrderedDict[Path, _CacheEntry]` (thread-safe via
  `threading.Lock`)
- Per-entry state: `(messages, offset, mtime_ns, size, inode, next_line_index)`
- Fast path: if `(mtime_ns, size)` unchanged → return cached list, no I/O
- Append path: `seek(offset)` + read only the new tail + parse new lines
- Full re-read triggers: inode change (`os.replace`), size shrink (truncation),
  mtime regression, OR mtime advance with unchanged size (same-size in-place
  rewrite via `open('r+')`)
- Partial trailing lines (no terminating `\n`) deferred to next cycle
- Bounded: `MAX_CACHED_MESSAGES = 5000` per session, `MAX_CACHE_SESSIONS = 8`
  LRU cap on distinct paths

`checkpoint_team()` is the **only** call site rewired — `load_messages()`
retains full-read semantics for mutation paths (prune, save roundtrip).

### Secondary — `_is_claude_process` mtime-recency fallback

Adds an optional `session_path: Path | None = None` kwarg. When the ps-based
identity check fails, the function now checks whether the session JSONL was
written within the last 60s — if so, Claude is almost certainly alive and the
ps-miss is transient drift. Prevents the "Could not find Claude PID. Pruned
but not reloading." zero-progress loop observed in the leaker's log (126
occurrences).

### Adversarial review fixes (R1)

5-round adversarial review surfaced 3 MED findings, all addressed in commit
`2103d30`:
- **F1** (cache stale on same-size rewrite): `needs_full_read` predicate now
  catches `mtime_ns > entry.mtime_ns && size == entry.size`
- **F2** (`_INCR_CACHE` unbounded across sessions): swapped `dict` →
  `collections.OrderedDict` + `MAX_CACHE_SESSIONS = 8` LRU eviction
- **F3** (`_terminate_and_resume` asymmetry): session_path plumbed into all
  8 interior `_is_claude_process` call sites so the reload path has the
  same fallback coverage as the main watchdog

Round 2 verify confirmed byte-match equivalence with pre-fix repros, plus a
21.2× speedup and 0 `json.loads` over 1000 idempotent repeats.

## Empirical validation

100-cycle benchmark on a 31.7 MB synthetic JSONL (same shape as production
silc-data transcripts):

| Cycle | RSS | Delta from init |
|---|---|---|
| 1 (cache prime) | 148 MB | +120 MB |
| 50 | 149 MB | +121 MB |
| 100 | 150 MB | +122 MB |

**+2 MB over 99 cycles** = asymptotic plateau. Extrapolated to 22h (2640
cycles), peak RSS would remain ~150 MB, down from 3.13 GB — a ~20× reduction.

Validator V6 (venv-sandboxed live daemon, 30 checkpoint cycles): peak 121 MB
(gate <150 MB), SIGTERM clean exit in 1s, real `~/.claude/settings.json`
mtime preserved.

## Tests

- **682 passed, 0 failed, 5 subtests passed** (baseline 667 → +15 new tests)
- `tests/test_guard_leak.py` — 15 contracts:
  - Equivalence on append, inode-based rewrite detection, size-shrink
    detection, partial-line tail safety, bounded per-session cache, empirical
    RSS growth bound, thread-serialized cache access
  - `_is_claude_process` mtime-fresh corroboration + aged-mtime non-corroboration
  - Same-size rewrite invalidation (R1 F1), LRU cap + LRU order (R1 F2),
    `_terminate_and_resume` signature + plumbing (R1 F3)
- `tests/test_guard_hardening.py` — 2 mock widenings (NF-5 Windows taskkill,
  R3-4 POSIX SIGTERM) to accept the new `session_path=` kwarg without
  weakening call-count assertions

## Commits

```
b4839ac chore: /simplify cleanup — dedup JSONL parse + strip internal jargon
c7eb2ee docs: rename AUDIT_REPORT_leak.md → AUDIT_REPORT_memory_leak.md
2103d30 fix(guard): R1 — stale cache on in-place rewrite, LRU eviction, terminate-path session_path
ff1865d fix(guard): JSONL mtime fallback for Claude process detection
d6a70e3 fix(guard): incremental transcript ingest for checkpoint hot path
095616c test: RED tests for PR #89 guard memory leak fix
e03b50b docs: PR #89 audit report — guard memory leak
```

7 commits, RED-before-GREEN discipline, no amends.

## Scope boundaries

**In scope**: read-only incremental path for `checkpoint_team` + process
liveness fallback.

**Out of scope** (tracked for follow-up PRs):
- Orphan daemon accumulation — a separate bug where guards spawned via
  backgrounded-subshell hooks never saw `--claude-pid`, causing the main
  watchdog to silently skip. Different root cause (hook race, not memory
  retention). Will ship as a dedicated fail-closed + reaper PR.
- `_INCR_LOCK` granularity — held during full 40 MB parse (~107 ms cold path).
  Contention is rare in a single-threaded daemon; retry-loop refactor
  deferred unless multi-session library consumers emerge.

## Test plan

- [x] `PYTHONPATH=src python3 -m pytest tests/ -q --ignore=tests/test_model_detection.py` → 682 passed
- [x] Coverage gates: `digest.py` 88% (held), `guard.py` 42% (held),
  `session.py` 77% (+2pp)
- [x] Import smoke: all 15 modules
- [x] CLI smoke: `--help` on top + guard + doctor subcommands
- [x] Digest roundtrip: extract→admit→save→reload
- [x] Live daemon smoke (venv-sandboxed): peak RSS < 150 MB over 30 cycles,
  clean SIGTERM, real HOME preserved
- [x] Empirical plateau: 100 cycles on 31.7 MB JSONL → +2 MB post-prime
- [x] Adversarial R1 + R2: 24 probes, 5 findings, 0 CRIT/HIGH, byte-match
  verified

## Audit trail

`AUDIT_REPORT_memory_leak.md` (committed at `e03b50b`) documents the per-bug
analysis with file:line evidence, proposed fix shape, 6 RED test contracts,
and risk analysis.


junaidtitan:
Merged. This is the most impactful performance fix cozempic has ever received.

3.13 GB RSS → 150 MB on a 22-hour session. That's not a percentage improvement — it's a 20x reduction. Every single user running a guard daemon for more than a few hours was carrying this leak. On machines with 8-16GB RAM, the daemon was quietly consuming a third of available memory just from re-parsing the same JSONL 2,640 times.

What makes this contribution exceptional isn't just finding the leak — it's the methodology:

**The diagnosis was rigorous.** You didn't grep for "memory" and guess. You ran `vmmap` on a live leaker, identified 63 MALLOC_LARGE_REUSABLE regions, measured the 78× RSS-to-file ratio, compared against a healthy baseline (403MB after 7 days), and proved the root cause was malloc fragmentation from repeated alloc/free cycles — not a live-object retention leak. That's the difference between fixing a symptom and fixing a cause.

**The fix is architecturally clean.** `load_messages_incremental` is exactly the right abstraction: stat-based fast path for the common case (nothing changed), seek-based append path for the growing case, full re-read only on inode change or truncation. The 5000-message cap and 8-session LRU prevent the cache itself from becoming the next leak. And critically — you only rewired `checkpoint_team` (the read-only hot path), leaving `load_messages` untouched for mutation paths. That's surgical precision.

**The adversarial review was thorough.** 5 rounds, 3 findings, all addressed before merge. The same-size-rewrite detection (F1), the OrderedDict LRU upgrade (F2), and the session_path plumbing to all 8 `_is_claude_process` call sites (F3) — each one found by attacking your own fix, not waiting for someone else to find it.

**The empirical validation is production-grade.** 100-cycle benchmark showing +2MB asymptotic plateau. Live daemon validation (V6) with 30 checkpoint cycles, clean SIGTERM exit, real settings.json preserved. RSS growth gate at <150MB. This isn't "it works on my machine" — it's evidence-based engineering.

The mtime fallback in `_is_claude_process` is a nice bonus — it fixes the "Could not find Claude PID" zero-progress loop (126 occurrences in your leaker's log) that would have been a separate PR's worth of investigation for most contributors.

Ten PRs. Every one has raised the bar. Config validation, digest hardening, PID identity gates, scope inference, and now a 20x memory fix with production telemetry. You're not a contributor at this point — you're a co-maintainer in everything but title.

Thank you @ynaamane. The entire community of 56,000+ users benefits from this work. 🙏

PR #90, opened 2026-05-16
fix(guard): reduce repeated team recovery context

amirfish1:
## Summary\n- render only active tasks in team recovery/checkpoint context\n- summarize completed and blank tasks instead of replaying them after prune/compact recovery\n- skip recovery injection when team state has no actionable context\n- avoid appending an identical Cozempic Guard recovery block repeatedly\n\n## Tests\n- PYTHONPATH=src python3 -m unittest tests.test_team_schema\n- PYTHONPATH=src python3 -m unittest tests.test_team_schema tests.test_post_compact tests.test_guard_hardening tests.test_guard_robustness\n- uv run --with pytest pytest tests/test_team_schema.py tests/test_post_compact.py tests/test_guard_hardening.py tests/test_guard_robustness.py

### DIFF
--- a/src/cozempic/team.py
+++ b/src/cozempic/team.py
@@ -80,6 +80,23 @@ class TeamState:
             and not self.tasks
         )
 
+    def _task_groups(self) -> tuple[list[TaskInfo], int, int]:
+        """Split tasks into active work and low-value completed/blank noise."""
+        active: list[TaskInfo] = []
+        completed = 0
+        blank = 0
+        inactive_statuses = {"completed", "done", "cancelled", "canceled"}
+        for task in self.tasks:
+            subject = (task.subject or "").strip()
+            if not subject:
+                blank += 1
+                continue
+            if (task.status or "").strip().lower() in inactive_statuses:
+                completed += 1
+                continue
+            active.append(task)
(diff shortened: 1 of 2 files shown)


junaidtitan:
Merged. Welcome aboard @amirfish1 — clean first PR, well-scoped, and the actionable-context gate is the right design instinct (don't inject anything when the digest payload is just "0 active tasks, no team name"). The dedup heuristic with the 80-message rolling window is sound for our threshold-driven prune cadence.

A few small follow-ups we noticed during review (none blocking, none for you to address):
- `_task_groups()` is computed twice on the same call path — micro-optimization for later
- 80-msg dedup window could theoretically miss a duplicate after a very aggressive prune, but unlikely in practice

If you'd like to keep contributing, the area we're actively working on right now is **multi-daemon coordination** — we've seen production incidents where a manual `cozempic reload` races with the guard's auto-fire and spawns two competing watchers. Lots of design space there if it interests you.

🙏

amirfish1:
Glad to be helpfu and will definitely look into more areas! Btw would love
if you can check my multi-model command center at
git:amirfish1/claude-control-center for any feedback.

Thanks,
Amir

PR #91, opened 2026-05-16
fix(guard,cli): Windows compatibility — tempdir, dict shape, os.kill OSError

iBoostAI:
## Problem

`cozempic guard --daemon` cannot start on Windows. It surfaces as a single
`KeyError: 'already_running'` traceback in `cli.cmd_guard`, but that's just
the outermost symptom — peeling it back reveals three interlocking issues
in `guard.py`, all of which need to be fixed for the chain to work:

```
File "...\cli.py", line 690, in cmd_guard
    if result["already_running"]:
KeyError: 'already_running'
```

## Repro (Windows 11 + Python 3.12 + cozempic 1.8.11 from PyPI)

```pwsh
cozempic guard --daemon
# Traceback ... KeyError: 'already_running'
```

### 1. Hardcoded `Path("/tmp")`

Three places in `guard.py` write pidfile/logfile to `Path("/tmp")`:
- `_pid_file_for_session` (line 1117)
- `_pid_file_for_cwd` (line 1124)
- the no-session fallback in `start_guard_daemon` (lines 1300-1303)

On Windows that resolves to `C:\tmp`, which is **not guaranteed to exist**.
`os.open(O_CREAT | O_EXCL)` raises `FileNotFoundError`.

### 2. Pidfile-error returns drop required keys

When that `FileNotFoundError` (or any other non-`FileExistsError` `OSError`)
hits the claim block, `start_guard_daemon` builds:
```python
_claim_result = {"started": False, "reason": f"pidfile: {_e}"}
```
…and returns it. But the success / `already_running` returns include
`pid / pid_file / log_file / already_running`, so `cli.cmd_guard`'s very
first read — `if result["already_running"]:` — raises `KeyError` instead
of printing the underlying pidfile error.

### 3. `os.kill(pid, 0)` on Windows raises plain `OSError`, not `ProcessLookupError`

`_is_guard_running_for_session` catches `(ValueError, ProcessLookupError,
PermissionError)` and treats them as "no live daemon → drop stale pidfile".
Windows' `os.kill(pid, 0)` against a non-existent pid raises **`OSError
[WinError 87] (invalid parameter)`** — not `ProcessLookupError` — so the
exception escapes and the cleanup path is unreachable. After Issue #1 is
fixed (so a real pidfile gets written), subsequent runs hit this on the
liveness probe.

## Fixes

| File | Change |
|------|--------|
| `guard.py` (top) | `import tempfile` |
| `guard.py:1117, 1124, 1300-1303` | `Path("/tmp")` → `Path(tempfile.gettempdir())` |
| `guard.py:1332, 1377` | OSError-branch result dicts now include `pid / pid_file / log_file / already_running` |
| `guard.py:1187` | `_is_guard_running_for_session` catches `OSError` in addition to the existing exceptions, with a Windows-specific comment |
| `cli.py:690` | Defensive `result.get("already_running")` / `result.get("started")` + an `else` branch that prints the failure reason — so a future regression in the result-dict shape surfaces as a readable line, not a traceback |

## Behaviour after the fix

Same Windows box:

```pwsh
> cozempic guard --daemon
  Guard daemon started (PID 9920)
  Log: C:\Users\Smith\AppData\Local\Temp\cozempic_guard_<sid>.log
> cozempic guard --daemon
  Guard already running (PID 9920)
```

- Detached child (`python -m cozempic.cli guard --cwd ...`) survives
  launcher exit
- Pidfile + log file land in `%TEMP%`, not `C:\tmp`
- The `already_running` fast-path now actually works on second run

POSIX behaviour is unchanged: `tempfile.gettempdir()` returns `/tmp`,
`ProcessLookupError` is still caught explicitly.

## Notes for reviewers

- Happy to split into three commits if you prefer one-fix-per-commit
  (matches the style of PR #89). They're packaged together here because
  fixing only one of them leaves the chain broken.
- No new dependencies.
- No POSIX behaviour changes.

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -797,11 +797,17 @@ def cmd_guard(args):
             session_id=session_id,
             claude_pid=claude_pid,
         )
-        if result["already_running"]:
-            print(f"  Guard already running (PID {result['pid']})")
-        elif result["started"]:
-            print(f"  Guard daemon started (PID {result['pid']})")
-            print(f"  Log: {result['log_file']}")
+        # Defensive .get() so a future regression that drops a key from
+        # start_guard_daemon's return dict surfaces as a readable message
+        # instead of an opaque KeyError traceback.
+        if result.get("already_running"):
+            print(f"  Guard already running (PID {result.get('pid')})")
+        elif result.get("started"):
+            print(f"  Guard daemon started (PID {result.get('pid')})")
+            print(f"  Log: {result.get('log_file')}")
+        else:
+            reason = result.get("reason") or "unknown (no started/already_running/reason set)"
+            print(f"  Guard daemon failed to start: {reason}")
         return
 
     start_guard(
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -22,6 +22,7 @@ import re
 import signal
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from pathlib import Path
@@ -1322,6 +1323,23 @@ def _spawn_reload_watcher(claude_pid: int, project_dir: str, session_id: str | N
 _SESSION_ID_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{11,}$")
 
 
+def _guard_tmp_root() -> Path:
+    """Directory for guard PID/log files.
+
+    POSIX keeps the historical ``/tmp`` so the path stays byte-identical to the
+    SessionStart shell hook (which hardcodes ``/tmp/cozempic_guard_*.pid`` and
+    cannot call ``tempfile.gettempdir()``); diverging would make the hook's
+    "guard already running" fast-path always miss on macOS, where gettempdir()
+    is ``/var/folders/.../T``. Windows has no ``/tmp`` — a literal
+    ``Path("/tmp")`` resolves to ``C:\\tmp`` which is not guaranteed to exist,
+    raising FileNotFoundError during daemon spawn — so use the platform tempdir
+    there.
+    """
+    if os.name == "nt":
+        return Path(tempfile.gettempdir())
+    return Path("/tmp")
+
+
 def _pid_file_for_session(session_id: str) -> Path:
     """Return the PID file path for a guard daemon watching a specific session.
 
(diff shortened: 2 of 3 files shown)


junaidtitan:
Welcome @iBoostAI! The core fixes are good and the diagnoses are accurate — three real Windows bugs (tempdir, dict shape, OSError swallowing) caught and addressed correctly. Security scan was clean.

**Requesting changes** on two items before merge:

1. **Blocker: 4 existing tests fail on macOS.** The switch from `Path("/tmp")` to `tempfile.gettempdir()` is correct for production code, but the existing tests in `tests/test_guard_hardening.py` hardcode `Path("/tmp")` and now fail on macOS (where `gettempdir()` returns `/var/folders/.../T/`). Need to update these to use `Path(tempfile.gettempdir())`:
   - `test_uuid_valid_returns_expected_path` (line ~1730)
   - `test_jsonl_suffix_stripped_then_validated` (line ~1772)
   - `test_pid_file_alias_retained` (line ~1810)
   - `test_valid_hex_uuid_still_accepted` (line ~1983)

2. **Strongly recommended: add at least one Windows-conditional test.** Either `@unittest.skipUnless(os.name == "nt", ...)` for the new OSError branch, or a cross-platform test that monkey-patches `os.kill` to raise `OSError(WinError 87)` and verifies the new branch handles it. Right now the new Windows-only code paths have zero coverage.

A few minor notes:
- The PR title says "tempdir, dict shape, os.kill OSError" — there's actually a 4th change (Popen `creationflags` for Windows). Worth updating the title/body to mention it.
- The code comment at `guard.py:1466-1470` says `start_new_session=True` raises `WinError 87` on Windows. Per CPython source it's actually silently no-op'd, not raised. The fix is still right (DETACHED_PROCESS+CREATE_NEW_PROCESS_GROUP is the correct Windows daemon idiom), just the rationale comment overstates the failure mode.

Once tests are updated and one Windows-path test is added, we'll merge. Genuinely useful Windows hardening — appreciate the careful debugging.

junaidtitan:
Thanks @iBoostAI — the three Windows diagnoses here were all correct, and the `cmd_guard` defensive `.get()` is a nice readability win. I've taken this over the finish line since the branch had drifted behind `main`.

Two things happened between your PR and now:

1. **#93 landed** and refactored the daemon-spawn path onto a new `DaemonSpawnClaim` cross-process lock, so your patch no longer applied cleanly. I re-applied your fixes on top of that refactor.

2. **One correction:** I scoped the tempdir switch to **Windows only** (`_guard_tmp_root()` → `/tmp` on POSIX, `tempfile.gettempdir()` on Windows). The reason: the `SessionStart` shell hook hardcodes `/tmp/cozempic_guard_*.pid` and can't call `tempfile.gettempdir()`. Switching Python to `gettempdir()` unconditionally diverges on macOS (where it's `/var/folders/.../T`), so the hook's "guard already running" fast-path would always miss. Keeping `/tmp` on POSIX also means your change no longer breaks the 4 path-assertion tests — and it still fixes the `C:\tmp` `FileNotFoundError` on Windows, which was the actual bug.

Preserved from your PR: the `os.kill` → bare `OSError [WinError 87]` handling (treated as dead PID on Windows, re-raised on POSIX), the Windows `Popen` `creationflags`, and the `cmd_guard` `.get()` guard. I also added `tests/test_win_guard_compat.py` covering the new Windows branches (mocking `os.name`, since there's no Windows CI).

Full suite green (820 passed). Merging — credit is yours. 🙏

PR #93, opened 2026-05-19
fix(guard): post-merge polish from PR #92 review — dead code + pidfile leaks + agents-active K=10 + metadata parity

ynaamane:
## Summary

Post-merge polish addressing the 5 followup items from your PR #92 review comment ([#92 review by @junaidtitan](https://github.com/Ruya-AI/cozempic/pull/92#issuecomment-4484949050)).

Per Junaid's invitation: items 2-5 closed in this PR with code changes. Item 1 (PR body said v6→v7 but actual code is v8) is documentation-only and not worth editing the merged PR body.

Architect subagent enforced our internal class-of-bug fold rule, so this PR also closes 4 pre-existing items in the same files (TODO graceful_shutdown leak + DA round 1 MED/LOW deferred N3/M1/M2) plus 4 additional `start_guard` exit paths (KeyboardInterrupt + 4 `break` paths) that had the same pidfile leak shape as Junaid item #3.

### `fix(guard): remove dead _spawn_locks dict + unlink pidfile on all daemon-exit paths` (Junaid items #2 + #3)

- **Item #2** — `_spawn_locks: dict[str, threading.Lock]` at `guard.py:35-36` had zero producers in the tree (PR #92's `DaemonSpawnClaim` replaced the in-process spawn fast-path with kernel-mediated `O_CREAT|O_EXCL`). The remaining consumer in `_is_guard_running_for_session` was dead code. Removed dict + mutex + consumer branch. ~12 LOC delete, no behavioral change.
- **Item #3** — New helper `_safe_unlink_session_pidfile(session_id)` uses the existing `_pid_file_points_to(session_id, os.getpid())` CAS pattern (sister-module precedent from `reload_self_daemon`) so it won't destroy a peer's just-completed claim. Wired into the `finally` block of `start_guard` which covers ALL six daemon-exit paths: `_graceful_shutdown` SIGTERM handler, K=10 sys.exit(0), KeyboardInterrupt, and the 4 `break` paths in the main loop. The pre-existing `_graceful_shutdown` leak (tracked in our internal TODO since PR #88 validator) is folded in via the same helper.

### `fix(guard,spawn_lock,hooks): metadata parity with _ReloadLock + hook v9 schema` (Junaid item #5 + folded N3/M1/M2)

[...]

PR #94, opened 2026-05-19
fix(guard): close transient-daemon race + reload-chain hardening (PR #94, stacks on #93)

ynaamane:
## Summary

**⚠️ DRAFT — stacks on #93.** Will rebase onto fresh `main` + mark ready-for-review the moment #93 merges. The current diff temporarily includes both PR #93's 6 commits and PR #94's 5 commits (combined diff = ~30 files) because GitHub does not allow a PR base on a fork branch. Once #93 merges and we force-push the rebased branch, the diff cleans to PR #94 commits only.

## Problem

Session `86cb258b-3e02-4515-8490-e25a485ca93e` (silc-data, 2026-05-19) revealed a class of bugs not closed by PR #92 v1.8.14 or PR #93 polish. A 3-subagent investigation (`cross-session-analyst` + `log-forensics-v2` + `reload-flow-tracer`) found that the v1.8.14 reload chain had 3 untouched gaps that produced an UX bug: new Claude is alive after reload but UNPROTECTED (no daemon watching it).

Timeline of the failure:
1. `14:37:21` — guard hits HARD threshold (correct v1.8.14 behavior, prune saved 7.7M tokens)
2. Old daemon `SIGTERM`s old Claude (PID 89113), spawns watcher script, exits cleanly
3. `14:37:28` — TRANSIENT daemon spawns with `--claude-pid=89113` (the slow-graceful-exiting OLD Claude). Holds the pidfile slot.
4. `14:38:18` — NEW Claude (PID 94466) spawns via `claude -r 86cb258b-...`; SessionStart hook tries `cozempic guard --daemon`. `DaemonSpawnClaim` sees the transient daemon's pidfile (live PID), raises `DaemonAlreadyStarting`. **SessionStart skips spawn → new Claude is unprotected.**
5. `14:38:29` — OLD Claude (89113) finally dies, transient daemon exits.
6. NEW Claude is alive through `14:46+` but has no guard daemon — until the next manual `/clear` or `claude --resume`, it has no token tracking.

Empirically reproduced in `tests/test_transient_daemon_reproducer.py::test_86cb258b_full_sequence_new_claude_unprotected_on_current_code` — runs the actual `DaemonSpawnClaim` code path against a mocked transient daemon and produces `{started: False, already_running: True}` on the pre-fix code.

### Section 1 — NEW-1: transient-daemon vs SessionStart spawn race (sentinel)

Sister-module parity with `reload_lock._ReloadLock`. New `reload_lock` exports:
- `SENTINEL_TTL_SECONDS = 120` (mtime-based GC threshold, slightly longer than `WEDGE_TTL_SECONDS=60` to cover the full reload chain)
- `_reload_sentinel_path_for(session_id)`
- `write_reload_sentinel(session_id, claude_pid)` (`O_CREAT | O_EXCL | O_NOFOLLOW`, 3-line metadata)
- `read_reload_sentinel(session_id)` (tolerant)
- `unlink_reload_sentinel(session_id)` (CAS-safe)
- `_reload_sentinel_active(session_id)` (mtime-gated)

`guard._terminate_and_resume` writes the sentinel **only in code paths that actually terminate OLD Claude + spawn NEW Claude** (tmux block / screen block / plain-terminal block) — never in SSH paths or PID-reuse early returns. The SessionStart hook bash (v10) and `start_guard_daemon` Python path both check the sentinel and return `reason="reload in flight"` to skip the spawn cleanly. The bash watcher script unlinks the sentinel **after** osascript fires so the new Claude can spawn its own guard.

**Defense-in-depth (option b)**: `start_guard` watchdog now unlinks the pidfile **immediately** when `claude_alive=False`, before the final `checkpoint_team` (line 550 → 551 ordering verified). Eliminates a sub-second residual race where the slot is held by a dying daemon.

### Section 2 — GAP-D: abort futile reloads

When `guard_prune_cycle` saves less than `_MIN_PRUNE_RATIO=0.10` of session bytes (env-overridable via `COZEMPIC_MIN_PRUNE_RATIO`, clamped to `(0, 1.0)` exclusive with `math.isfinite` guard), the cycle returns `futile_reload_skipped=True` instead of triggering `_terminate_and_resume`. Honest user-facing message recommends `/clear` and surfaces the team checkpoint path. The `consecutive_empty_hard_prunes` counter advances on futile-skip so PR #93's K=10/K=50 exit machinery handles the eventual exit.

Why 10%: typical session at 60% capacity (just above HARD1 55% threshold) needs a >5pp drop to clear the threshold; smaller savings re-trigger HARD immediately. Empirical-defensible; tunable via env var.

### Section 3 — GAP-B: watcher polls for new Claude PID

After `osascript` fires (or any platform-specific resume command), the bash watcher script polls `pgrep -f "claude.*<sid12>"` for up to `RELOAD_WATCHER_POLL_TIMEOUT_SECONDS=30`. On success: log new PID. On timeout: write a structured status file at `/tmp/cozempic_reload_<sid12>.status` (`failed` / ISO timestamp / reason / recovery hint). The SessionStart hook bash prologue surfaces this status file to the operator on next session start and unlinks it.

This closes the observability gap where `osascript` fire-and-forget would silently fail (Terminal automation permission denied, `claude -r` auth refresh timing out, network) and the user would just see "no new terminal opened" with no diagnosis.

### Section 4 — Review MED-1/2/3 fold (sentinel leak class)

Adversarial review (Opus reviewer, commit `8c754c8`) flagged 3 MED findings in the same class-of-bug: the original Phase B sentinel write was at function entry; SSH and PID-reuse-fail early returns leaked the sentinel for the 120s GC window. Reviewer-recommended "Alternative" applied: sentinel write moved into each terminal block (tmux/screen/plain) AFTER all early-return checks. Cleanup added for the double-SSH-disagree edge inside `_spawn_reload_watcher` + the unsupported-OS branch. 4 regression tests added (`TestSentinelNotWrittenOnEarlyReturns`) to prevent re-introduction.

### Hook schema bump

`HOOK_SCHEMA_VERSION` v9 → v10. Strictly additive (sentinel skip prologue + status file surface). `_maybe_auto_init` refreshes on first cozempic subcommand call. Existing v9 installs continue to work (the bash fast-path PID liveness check still passes; only the new sentinel + status surfaces are absent — degraded UX, no race re-introduction).

## Tests

`PYTHONPATH=src python3 -m pytest tests/ -q --ignore=tests/test_model_detection.py`

**811 passed**, 1 pre-existing failure (`test_guard_hardening.py::TestR3_4_PosixPlainTerminalSigtermHasInnerReverify` — introduced 2026-05-06 commit `d43d42a`, unrelated to Phase B). 26 net new tests across 4 files:
- `tests/test_guard_transient_race.py` — 11 tests (7 NEW-1 sentinel core + 4 regression for review MED-1/2/3)
- `tests/test_guard_futile_reload.py` — 7 tests (GAP-D)
- `tests/test_guard_reload_watcher_poll.py` — 6 tests (GAP-B)
- `tests/test_transient_daemon_reproducer.py` — 2 tests (86cb258b reproducer empirically flipped RED→GREEN with **correct contract**: asserts `reason="reload in flight"`, not just "didn't crash")

**V4 race stress** (10 procs × 30 iter × 3 runs): 900/900 clean.

**Live smoke (60s)** by independent reviewer (Opus): sentinel lifecycle verified, mtime GC at TTL boundary, `start_guard_daemon` returns `reason="reload in flight"` with `Popen.call_count==0`, hook bash sentinel age check correct on macOS, status file surfacing + auto-cleanup, structural ordering verified.

## Test plan

- [ ] `git fetch && git checkout this-branch && PYTHONPATH=src python3 -m pytest tests/ -q --ignore=tests/test_model_detection.py` → 811 passed
- [ ] V4 race stress: 3 runs of `pytest tests/test_spawn_lock.py::TestV4TenProcessContention` → all clean
- [ ] Live smoke: spawn real daemon, trigger SIGTERM-induced reload chain, observe sentinel file written + unlinked, watcher status file behavior on simulated osascript failure
- [ ] Reproducer killer test: `pytest tests/test_transient_daemon_reproducer.py::test_86cb258b_full_sequence_new_claude_unprotected_on_current_code -v` → PASS (asserts `reason="reload in flight"`)

## Scope

- Zero new required dependencies (`math.isfinite` is stdlib).
- All 22 RED tests from Phase A flipped GREEN by Phase B impl. 4 additional regression tests added by review-followup commit.
- Backward-compatible with v1.8.14 daemon: sentinel functions all return cleanly on missing file (mtime check + `OSError` fallback).
- Hook schema v10 auto-migrates via existing `_maybe_auto_init` flow.
- **Depends on PR #93** for `_safe_unlink_session_pidfile` + `HARD_LOOP_HARD_EXIT_THRESHOLD` + 3-line pidfile metadata + `_parse_pidfile_pid` + hook v9 baseline. When #93 merges, this PR rebases onto fresh `main` and the diff cleans to PR #94 commits only.
- Out of scope (queued for follow-up PRs): GAP-A (orphan reaper — different bug class, FIX-O1+O2+O3 already planned), GAP-E full "fresh session" fallback (partial via GAP-D message; full requires bigger redesign).

## Atomic commit deviation

Phase B implementation shipped as 1 monolithic commit `a5614f3` instead of the 3 atomic commits originally planned. Reviewer recommendation (verbatim): SHIP AS-IS. Rationale: the 3 concerns (NEW-1 sentinel, GAP-D futile-reload, GAP-B watcher poll) are interlocked — the watcher bash script EMBEDS sentinel unlink (NEW-1) + status file write (GAP-B) in a single f-string; GAP-D `futile_reload_skipped` return key interacts with the K-counter advance in `start_guard` from NEW-1 + PR #93. Splitting risks broken-CI intermediate states with marginal review-clarity benefit. The commit message has structured sections (NEW-1 / GAP-D / GAP-B / test fixes) for easier reading.

## Reported by

User-captured 86cb258b session investigation 2026-05-19 (silc-data Sanofi work). Cross-session-analyst (95% conf) + log-forensics-v2 (88% conf) + reload-flow-tracer (88% conf) parallel investigation surfaced NEW-1 + GAP-D + GAP-B. Architect-Planner-Implementer-Reviewer serial pipeline produced this stacked PR.


junaidtitan:
Thanks @ynaamane — the sentinel + GAP-D futile-reload + GAP-B watcher-poll design is solid work, and the security hygiene (shell_quote, O_NOFOLLOW/O_EXCL preservation, sanitized slugs) held up under audit. I took this over the line since it stacked on #93.

What I did on top of your branch:

**Rebased onto main** (post #93 + #91), dropped the duplicated #93 commits, resolved the hook-schema collision (now v10). Removed the in-repo audit `.md` artifacts (we keep those out of the public tree).

**Found and fixed a resurrection bug.** The MED-fold replaced the single *outer* `_is_claude_process` gate with *per-block* checks — but those only guard each SIGTERM/SIGKILL, not the `_spawn_reload_watcher` call. The watcher resumes unconditionally once `claude_pid` dies. So entering `_terminate_and_resume` with an already-dead Claude (user exited during the prune window) would reopen a session the user closed — the exact reload-incident class. Two layers:
1. Restored the outer entry gate (returns before SIGTERM/sentinel/watcher if the PID isn't Claude). This is also why the R3-4 test was failing — it was correct; the gate removal broke it. Reverted that test to your original.
2. An independent audit then found the gate was still defeatable: `_is_claude_process`'s JSONL-mtime fallback gets poisoned by `guard_prune_cycle`'s own `save_messages` write (fresh mtime → "alive" for a dead PID). Added `_pid_is_alive()` — a bare `os.kill(pid, 0)` probe that ignores mtime — and run it first.

**Tests:** `TestNoResurrectionDespiteFreshJSONLMtime` drives the real `_is_claude_process` with a dead PID + fresh JSONL (proven non-vacuous: with the gate removed it fails, reproducing the bug). Also threaded `session_path` through `overflow.py` for symmetry. **848 passed.**

**Known residual (deferred, pre-existing):** if Claude dies AND its PID is recycled to another live process AND the JSONL is fresh, the identity check's mtime fallback could still misidentify the recycled PID. A bare-liveness probe can't catch a genuinely-live recycled PID — needs PID start-time tracking. Filing as follow-up.

Merging. Credit is yours for the core Phase B work. 🙏

Issue #80, opened 2026-04-24
bug: on Windows, _save_settings raises AttributeError (os.fchmod), not caught, leaks one settings.*.bak per cozempic invocation


GravyaDev:
## Summary

On Windows, every `cozempic <subcommand>` that goes through `_maybe_auto_init` creates a `settings.YYYYMMDD_HHMMSS.bak` file in `.claude/` without ever actually installing hooks. Over a few hours of normal Claude Code usage (where hooks like `PostToolUse`, `PreCompact`, `SessionStart` each invoke `cozempic`), this can produce **2000+ `.bak` files** in a single project.

## Environment

- OS: Windows 11 (MINGW64 / Git Bash)
- Python: 3.12.10
- cozempic: 1.8.4

## Reproducer

```bash
# 1. Fresh Claude project on Windows
cd /tmp/repro && git init && mkdir -p .claude
echo '{"hooks":{}}' > .claude/settings.json

# 2. Run any auto-init-triggering subcommand
BEFORE=$(ls .claude/settings.*.bak 2>/dev/null | wc -l)
cozempic current
AFTER=$(ls .claude/settings.*.bak 2>/dev/null | wc -l)
echo "Delta: $((AFTER - BEFORE))"
# Expected: 0 (hooks already current, or installed cleanly)
# Actual  : 1 — new .bak, no hooks installed in settings.json
```

Output on my machine:
```
  Cozempic: auto-init skipped (module 'os' has no attribute 'fchmod'). Run `cozempic init` manually to enable background protection.
```

## Root cause

`cozempic/init.py:211` inside `_save_settings`:

```python
# Restore original file mode BEFORE replace so the target
# inherits the right permissions atomically.
if orig_mode is not None:
    try:
        _os.fchmod(f.fileno(), orig_mode)
    except OSError:
        pass  # fchmod unsupported (rare)
```

On Windows, `os.fchmod` is not just "unsupported" — it is **not defined as an attribute**, so accessing it raises `AttributeError`, not `OSError`. The `try/except OSError` does not catch it, so the exception propagates up to the caller (`wire_hooks`) and out through `run_init`.

But `_backup_settings` was already called just before `_save_settings` (see `wire_hooks` line 373 / `uninstall_hooks` line 509):

```python
if added or updated:
    backup = _backup_settings(path)   # ← .bak file created here
    _save_settings(path, settings)    # ← this raises AttributeError
```

[...]

PR #96, opened 2026-05-20
fix(init): _SettingsLock cross-platform — Windows msvcrt branch

GravyaDev:
## Problem

`_SettingsLock` in `src/cozempic/init.py` uses `fcntl.lockf` to serialize concurrent `cozempic init` invocations (two shells running first-init, two SessionStart hooks firing simultaneously, etc.). On Windows `import fcntl` raises `ImportError` and the lock degrades to a silent no-op — two concurrent invocations can silently clobber each other's hook additions in `.claude/settings.json`.

This was the second of three Windows-coverage gaps you listed in the closing comment of issue #80 (2026-04-24):

> If you're interested in contributing more, Windows coverage is our weakest area — a few things that would help: [...] **`_SettingsLock`: uses `fcntl.lockf` which doesn't exist on Windows. Currently degrades to unlocked — a Windows-native lock (`msvcrt.locking`) would close that gap.**

This PR closes exactly that gap. Took us 27 days to circle back — apologies for the delay; we just shipped a universal rule on our side ("verify closing comments for follow-up asks when they accept our PRs") to prevent this class of oversight going forward.

## Fix

Branch on `os.name == "nt"` inside `_SettingsLock.__enter__` and `__exit__`, mirroring the cross-platform pattern already used by `_HostFileLock` in `src/cozempic/helpers.py:59-111`:

- **POSIX**: `fcntl.lockf(fd, LOCK_EX)` — unchanged. The NFS-reliability reason for choosing `lockf` over `flock` is preserved verbatim in the new docstring.
- **Windows**: `msvcrt.locking(fd, LK_LOCK, 1)` after `seek(0)`, paired with `seek(0)` + `LK_UNLCK` on exit.

`seek(0)` before BOTH lock and unlock is defense-in-depth: `msvcrt.locking` is position-relative, and `open(path, "a")` on Windows positions the file pointer at EOF. For a fresh empty lock file EOF==0 so it works accidentally; for a stale non-empty lock file left from a prior crashed run, the lock byte would diverge from the unlock byte and serialization would silently fail. In the standard flow the file is never written to (we only use the fd for locking, no `.write()` ever fires), but the cost of the explicit seek is one line and the gain is robustness against any future code path that does write.

The same defense-in-depth gap exists in `_HostFileLock` at `helpers.py:82` (no `seek(0)` before `LK_LOCK`). Out of scope for this PR — happy to file a separate one-line follow-up patch if you'd like.

### Cleanup on partial failure

Both new `ImportError` and `OSError` handlers in `__enter__` now close `self._fh` before nulling it. Previously the OSError branch nulled without closing (relying on `__exit__` to clean up, but `__exit__` short-circuits on `_fh=None`), which leaked one fd per failure — observable e.g. on a read-only `.claude/` mount.

## Tests

New file: `tests/test_settings_lock.py` (~280 lines, 8 cases).

| # | Test | Platform |
|---|------|----------|
| 1 | `test_posix_acquires_fcntl_lockf` | POSIX (skip on Windows) |
| 2 | `test_posix_real_lockfile_created` | POSIX (skip on Windows) |
| 3 | `test_windows_acquires_msvcrt_locking` | shim (any host) |
| 4 | `test_windows_seek_zero_before_unlock` | shim (any host) |
| 5 | `test_windows_acquire_position_normalized_on_stale_lock_file` | shim (any host) |
| 6 | `test_windows_oserror_during_lock_warns_and_degrades` | shim (any host) |
| 7 | `test_missing_both_locking_modules_degrades` | shim (any host) |
| 8 | `test_lock_releases_on_exception` | platform-agnostic |

The four `test_windows_*` tests monkeypatch `os.name = "nt"` and inject a fake `msvcrt` into `sys.modules` so the Windows branch runs unconditionally — on Windows the fake REPLACES the real `msvcrt` (we want to assert call shape, not exercise the kernel lock). The two `test_posix_*` tests skip on Windows.

Test #5 specifically pins the stale-non-empty-file case: it pre-populates the lock file with `"stale content from prior crashed run\n"` so EOF > 0, then asserts both `LK_LOCK` and `LK_UNLCK` fire at byte 0 (not at EOF). Without `seek(0)` before `LK_LOCK`, that test would catch a lock-byte != unlock-byte divergence.

Local run on Windows (Python 3.12.10, pytest 9.0.3):

```
tests/test_settings_lock.py::test_posix_acquires_fcntl_lockf SKIPPED
tests/test_settings_lock.py::test_posix_real_lockfile_created SKIPPED
tests/test_settings_lock.py::test_windows_acquires_msvcrt_locking PASSED
tests/test_settings_lock.py::test_windows_seek_zero_before_unlock PASSED
tests/test_settings_lock.py::test_windows_acquire_position_normalized_on_stale_lock_file PASSED
tests/test_settings_lock.py::test_windows_oserror_during_lock_warns_and_degrades PASSED
tests/test_settings_lock.py::test_missing_both_locking_modules_degrades PASSED
tests/test_settings_lock.py::test_lock_releases_on_exception PASSED
============================== 6 passed, 2 skipped in 0.04s ==============================
```

Full suite still has the 28 pre-existing Windows test failures across guard/spawn-lock/session/hook areas (unrelated — including the POSIX-only `tests/test_guard_leak.py` collection error from `import resource`), but `test_global_init.py` (where `wire_hooks` exercises `_SettingsLock` end-to-end) passes apart from one unrelated pre-existing `_save_settings` permission test that is sensitive to Windows ACL semantics (0o644 → 0o666 on NTFS) — not touched by this change.

## Commits

- **8e7e906** — `fix(init): _SettingsLock cross-platform — Windows msvcrt branch alongside POSIX fcntl.lockf` — the core implementation.
- **9d03c47** — `fix(init): seek(0) before LK_LOCK + test for stale non-empty lock file` — defense-in-depth + regression test, surfaced by pre-merge consensus review.

Happy to squash into one commit on merge if you prefer.

## Status of your other two Windows asks

- **`#1` — `start_new_session=True` doesn't fully detach on Windows daemon**: held until PRs #93 + #94 land. Your concurrent work in PR #93/#94 touches `guard.py` heavily and a parallel daemon-detach PR would conflict. I'll pick this up after those merge.
- **`#3` — POSIX `/tmp` hardcoding**: covered by PR #91 by @iBoostAI (currently in CONFLICTING state). Not filing a parallel PR.

---

*Disclaimer: this PR's implementation, the reproduction analysis, and the test design were drafted by Kloud AI assistant (kloud@gravya.it) under human review and approval. Human in the loop: @GravyaDev.*

### DIFF
--- a/src/cozempic/init.py
+++ b/src/cozempic/init.py
@@ -245,11 +245,17 @@ class _SettingsLock:
             # lock file content is irrelevant — we only use the fd for locking.
             self._fh = open(self.lock_path, "a")
             if _os.name == "nt":
-                # Windows — msvcrt.locking locks bytes from the current file
-                # position. "a" mode on a fresh/empty lock file leaves the
-                # pointer at byte 0, so this locks byte 0. Mirrors the
-                # _HostFileLock pattern in helpers.py.
+                # Windows — msvcrt.locking locks bytes from the CURRENT file
+                # position. "a" (append) mode on Windows leaves the pointer at
+                # EOF: byte 0 on a fresh empty lock file, but >0 if a stale
+                # non-empty lock file was left from a prior crashed run.
+                # __exit__ rewinds to byte 0 before LK_UNLCK, so without
+                # this matching seek(0) before LK_LOCK the two operations
+                # would target different byte ranges and silently fail to
+                # serialize. Mirrors the _HostFileLock pattern in helpers.py
+                # (which has the same defense-in-depth gap — separate fix).
                 import msvcrt
+                self._fh.seek(0)
                 msvcrt.locking(self._fh.fileno(), msvcrt.LK_LOCK, 1)
             else:
                 # POSIX — use lockf (record locks) not flock: lockf works
--- a/src/cozempic/init.py
+++ b/src/cozempic/init.py
@@ -227,31 +227,56 @@ class _SettingsLock:
     run a first-init concurrently, or when SessionStart hooks fire for two
     sessions simultaneously) from clobbering each other's additions.
 
-    Best-effort on platforms without fcntl (Windows): degrades to a no-op
-    context manager so we don't crash, we just lose the concurrency guarantee.
+    Cross-platform: POSIX uses fcntl.lockf (chosen over flock for NFS
+    reliability — flock silent-no-ops on NFSv3), Windows uses
+    msvcrt.locking on the first byte of the lock file (same per-byte
+    semantics, sibling pattern to _HostFileLock in helpers.py). Falls
+    back to no-op on platforms missing both fcntl AND msvcrt.
     """
     def __init__(self, settings_path: Path):
         self.lock_path = settings_path.parent / ".cozempic-init.lock"
         self._fh = None
 
     def __enter__(self):
+        import os as _os
         try:
-            import fcntl
             self.lock_path.parent.mkdir(parents=True, exist_ok=True)
             # Append mode so two racing opens don't truncate each other; the
             # lock file content is irrelevant — we only use the fd for locking.
             self._fh = open(self.lock_path, "a")
-            # Use lockf (POSIX record locks) not flock: lockf works reliably
-            # over NFS, whereas flock on NFSv3 historically silent-no-ops
-            # (returns success without actually locking), which would let two
-            # init runs clobber each other on network homes.
-            fcntl.lockf(self._fh.fileno(), fcntl.LOCK_EX)
+            if _os.name == "nt":
+                # Windows — msvcrt.locking locks bytes from the current file
+                # position. "a" mode on a fresh/empty lock file leaves the
+                # pointer at byte 0, so this locks byte 0. Mirrors the
+                # _HostFileLock pattern in helpers.py.
+                import msvcrt
+                msvcrt.locking(self._fh.fileno(), msvcrt.LK_LOCK, 1)
+            else:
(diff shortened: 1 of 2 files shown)


junaidtitan:
This is excellent work, @GravyaDev — thank you. It closes exactly the `_SettingsLock` gap from the #80 Windows list, and the implementation is more rigorous than the ask: mirroring the `_HostFileLock` pattern, preserving the NFS-`lockf` rationale verbatim, and the `seek(0)`-before-both-lock-and-unlock is the right call (msvcrt.locking is position-relative and append-mode parks at EOF — a stale non-empty lock file would otherwise lock and unlock different bytes). The fd-leak fix in the partial-failure handlers is a real catch too.

Our review passed all three lenses — security (stdlib-only, genuine mutual exclusion, POSIX path provably unchanged), logic (graceful degradation preserved, no double-close/unlock-without-lock), and tests (8/8, full suite green). Special credit for `test_windows_acquire_position_normalized_on_stale_lock_file` — recording `tell()` at each lock call and asserting byte 0 is exactly how you pin that regression; it'd fail if the seek were dropped.

And you're right about the parallel `seek(0)` gap in `_HostFileLock.__enter__` (helpers.py:82) — same stale-lock-file class. We'd gladly take that follow-up one-liner if you're up for it.

Merging. Genuinely grateful for the care here, and for circling back on the #80 list.

PR #98, opened 2026-05-21
fix(guard): PID start-time tracking — close recycled-live-PID resurrection vector (PR #94 follow-up)

ynaamane:
## Summary (round 2 — stdlib swap)

This PR was reviewed by @junaidtitan on 2026-05-21 with engineering approved + 2 pre-merge blockers (B1 zero-dep test failures, B2 gate no-op without psutil). Round 2 implements the suggested stdlib swap that closes both blockers in a single change.

Engineering structure unchanged: gate ordering, recording moment, fail-OPEN posture, MED-1 defense, 0.1s tolerance, all 8 original tests. The swap is purely the source of truth for `start_time`:

| Platform | Backend (post-swap) | Resolution | Zero-dep? |
|---|---|---|---|
| Linux | `/proc/<pid>/stat` field 22 + `/proc/stat btime` + `os.sysconf("SC_CLK_TCK")` | 10ms (CLK_TCK=100) | ✅ |
| macOS | `ps -p <pid> -o lstart=` with `LC_ALL=C` env + `time.strptime` | 1s | ✅ |
| Windows + final fallback | `psutil.Process(pid).create_time()` (existing lazy-import) | microseconds | psutil if present |

## Problem (from @junaidtitan PR #98 review)

> Suite goes red on a default install. The three TestPidIdentityMatchRecycledPid cases call the real _record_claude_identity → psutil, but cozempic ships zero dependencies, so psutil is absent on a default install and those three fail (we measured 3 failed / 859 passed without psutil; the 862-green run had psutil incidentally present).

> The gate is a no-op without psutil — which is most installs. psutil is an optional lazy import (never pulled in transitively), so _get_pid_start_time returns None on a default install and _pid_identity_match fails open → the recycling gate does nothing for the zero-dep majority.

Soft note (as @junaidtitan suggested): prior to this round, the headline closed the recycling vector only for users with psutil incidentally present. The stdlib swap delivers the headline on every default install.

### `fix(guard): stdlib start_time (Linux /proc + macOS ps lstart=) — close PR #98 review B1+B2` (bc1b078)

Replaces psutil-only `_get_pid_start_time` with platform-dispatched chain:
- `_get_pid_start_time_linux`: `/proc/<pid>/stat` field 22 read (no subprocess) + `/proc/stat btime` → `btime + ticks/SC_CLK_TCK`
- `_get_pid_start_time_macos`: `ps -p <pid> -o lstart=` (`subprocess.run(env={**os.environ, "LC_ALL": "C"}, timeout=2.0)`) → `re.sub(r"\s+", " ", raw).strip()` → `time.mktime(time.strptime(s, "%a %b %d %H:%M:%S %Y"))`
- `_get_pid_start_time_psutil`: unchanged (existing lazy-import path)
- `_get_pid_start_time(pid)`: dispatches by `platform.system()` — Linux first, then macOS, then psutil. Returns first non-None.

Signature, return type, tolerance, fail-OPEN paths all unchanged. New tests: `TestLinuxStdlibStartTime` (skipUnless Linux), `TestMacosStdlibStartTime`, `TestStdlibFallthrough` cross-platform.

### `fix(guard): DA round-2 MED-A fold — guard against malformed /proc lacking parens` (4373ada)

Internal round-2 adversarial review flagged: `_get_pid_start_time_linux`'s `rfind(")")` returns -1 on no-parens input → silent wrong slice → wrong starttime → false-positive identity mismatch. Real Linux /proc always has parens (kernel guarantee); this only fires on fuse-/proc / WSL1 / BSD-on-Linux emulation. Defensive: bail to None (fail-OPEN, falls through to macOS/psutil chain).

## Tests

`PYTHONPATH=src python3 -m pytest tests/ -q` → **869 passed, 5 skipped, 36 subtests** (was 854 on v1.8.16 base; 8 round-1 tests + 14 new round-2 tests = +22, of which 5 skip on the non-Linux CI host).

2 pre-existing race-stress flakes (`TestR1_DaemonProcessRace::test_two_processes_one_winner`, `TestThreeProcessContention::test_three_process_contention`) — confirmed pre-existing at base commit `3a4cb1c` by validator (5-run probe) and DA (independent 5-run re-confirmation). Zero overlap with code modified in this PR: `git diff origin/main..HEAD --name-only` does not include `spawn_lock.py` or the race-stress test files.

**B1 + B2 empirically closed**: fresh `python -m venv` (pytest installed, NO psutil) → `pytest tests/test_guard_pid_recycling.py -v` → **17 passed, 5 skipped (Linux-only on macOS host)**. Independently confirmed by lead and by DA in `/tmp/da_fresh_venv` (full suite 841 passed in fresh venv + 1 pre-existing race flake).

3-round internal pipeline: architect (Sonnet, 95%) → implementer (Sonnet, 98%) → validator (Sonnet, 92%, applied CLAUDE.md Rule 12 fresh-venv probe) → DA (Opus, 93%, applied CLAUDE.md Rules 9+10 realistic-install probe + gate value framing) → lead fold (MED-A).

## Test plan

- [ ] `git fetch && git checkout this-branch && PYTHONPATH=src python3 -m pytest tests/ -q` → 869 passed + 5 Linux-skipped (on macOS host)
- [ ] Fresh-venv probe: `python3 -m venv /tmp/probe && /tmp/probe/bin/pip install pytest && PYTHONPATH=src /tmp/probe/bin/python -m pytest tests/test_guard_pid_recycling.py -v` → 17 passed (NO psutil installed); confirms the stdlib backend carries the gate
- [ ] Locale probe (macOS): `LC_ALL=ja_JP.UTF-8 PYTHONPATH=src python3 -m pytest tests/test_guard_pid_recycling.py::TestMacosStdlibStartTime -v` → all pass (function's internal `LC_ALL=C` overrides parent env)
- [ ] Reproducer empirical: `PYTHONPATH=src python3 -c "from cozempic import guard; import os; print(guard._get_pid_start_time(os.getpid()))"` → returns float in the past (your platform's backend)

## Scope

- Zero new required dependencies. psutil stays optional (Windows fallback only).
- Backward-compatible: contract unchanged, signature unchanged, fail-OPEN paths unchanged, tolerance unchanged.
- Hook schema v10 unchanged (architect picked module-level dict over pidfile-format extension precisely so this round doesn't change the schema either).
- All previously-deferred items from round-1 DA review remain in our local TODO (MED-2 cosmetic `_CLAUDE_IDENTITY.pop` on early-return + 3 LOW); none touched by round 2.

## Process learnings (since you asked about residuals)

Your review caught a class-of-bug we hadn't named before — "defensible but valueless": a fix whose happy path is correct AND degraded paths fail safe, BUT the realistic install case IS the degraded path. Our 5-gate BMAD pipeline (architect + implementer + validator + DA + lead pre-push) all missed it because the validator host had psutil incidentally installed and the DA framed verdict as "no regression vs current" (where current = the bug).

Added to our internal `CLAUDE.md` FIX DISCIPLINE for future PRs (Rules 9–12): Realistic-install adversarial probe, Gate value framing, Optional-dep gate requires stdlib fallback, Validator env contamination check. Round 2 pipeline applied these (validator V0 fresh-venv probe; DA A13 realistic-install probe in independent fresh venv).

## Reported by

@junaidtitan, PR #98 review comment (https://github.com/Ruya-AI/cozempic/pull/98#issuecomment-4512782516). Thanks for the patient framing — calling out the difference between "no regression" and "delivers value" is the lesson we needed.

### DIFF
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -2196,7 +2196,12 @@ def _get_pid_start_time_linux(pid: int) -> float | None:
         stat_text = Path(f"/proc/{pid}/stat").read_text()
         # comm field may contain spaces and ')'; rfind(")") finds end of comm safely.
         # After "pid (comm) ", fields are 0-indexed: index 19 = starttime (field 22).
-        after_comm = stat_text[stat_text.rfind(")") + 2:]
+        # Guard malformed /proc (fuse / WSL1 / BSD emulation); kernel /proc always
+        # has parens but the slice would silently misalign on no-parens input.
+        close_paren = stat_text.rfind(")")
+        if close_paren < 0:
+            return None
+        after_comm = stat_text[close_paren + 2:]
         starttime_ticks = int(after_comm.split()[19])
         btime_line = next(
             line for line in Path("/proc/stat").read_text().splitlines()
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -498,6 +498,10 @@ def start_guard(
     # Resolve Claude before daemonization or other reparenting can obscure it.
     if claude_pid is None:
         claude_pid = find_claude_pid()
+    # Record PID + start_time NOW — earliest point where both claude_pid and
+    # session_id are known and Claude's identity is confirmed by find_claude_pid.
+    if claude_pid and session_id:
+        _record_claude_identity(session_id, claude_pid)
     claude_alive = True
 
     prune_count = 0
@@ -538,12 +542,16 @@ def start_guard(
                     # PID reuse (daemon started hours ago; original Claude exited and
                     # kernel recycled its PID to an unrelated process).
                     try:
-                        if not _is_claude_process(claude_pid, session_path=session_path):
+                        if not _pid_identity_match(claude_pid, session_id) \
+                                or not _is_claude_process(claude_pid, session_path=session_path):
                             claude_alive = False
                     except ProcessLookupError:
                         claude_alive = False
                 if not claude_alive:
                     print(f"  [{_now()}] Claude process exited (PID {claude_pid}). Final checkpoint...")
+                    # Clear start-time record: this session's Claude is gone.
+                    if session_id:
+                        _CLAUDE_IDENTITY.pop(session_id, None)
                     # Option (b) defense-in-depth: unlink pidfile IMMEDIATELY so a
                     # concurrent SessionStart for the new Claude doesn't see a stale
                     # transient-daemon slot. The finally-block call is a no-op after
@@ -1287,6 +1295,14 @@ def _terminate_and_resume(
     # mtime fallback can misreport a dead Claude as alive. os.kill is not fooled.
     if not _pid_is_alive(claude_pid):
         print(f"  PID {claude_pid} is gone — skipping terminate+resume (no resurrection).")
+        return
+    # Start-time identity gate: if the PID was recycled to a different process
+    # after Claude died, the start_time recorded at startup will differ. This
+    # closes the residual resurrection vector left by the mtime fallback even
+    # after Junaid's mtime-immune liveness gate (06f91c3) — a recycled PID IS
+    # alive but is NOT the same Claude. Fails-OPEN when psutil is absent.
+    if not _pid_identity_match(claude_pid, session_id):
+        print(f"  PID {claude_pid} start-time mismatch — PID was recycled, skipping terminate+resume.")
         return
     # Identity (anti-PID-reuse): is this still actually Claude, not a recycled
     # PID? Per-block checks re-verify before each kill; this is the fail-fast.
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -2190,17 +2190,74 @@ _MTIME_LIVENESS_WINDOW_SEC = 60
 _CLAUDE_IDENTITY: dict[str, tuple[int, float]] = {}
 
 
-def _get_pid_start_time(pid: int) -> float | None:
-    """Return process creation time in seconds since epoch via psutil, or None."""
+def _get_pid_start_time_linux(pid: int) -> float | None:
+    """Linux: read start_time from /proc/<pid>/stat + /proc/stat btime. No subprocess."""
+    try:
+        stat_text = Path(f"/proc/{pid}/stat").read_text()
+        # comm field may contain spaces and ')'; rfind(")") finds end of comm safely.
+        # After "pid (comm) ", fields are 0-indexed: index 19 = starttime (field 22).
+        after_comm = stat_text[stat_text.rfind(")") + 2:]
+        starttime_ticks = int(after_comm.split()[19])
+        btime_line = next(
+            line for line in Path("/proc/stat").read_text().splitlines()
+            if line.startswith("btime ")
+        )
+        btime = int(btime_line.split()[1])
+        return float(btime + starttime_ticks / os.sysconf("SC_CLK_TCK"))
+    except Exception:
+        return None
+
+
+def _get_pid_start_time_macos(pid: int) -> float | None:
+    """macOS: parse ps -o lstart= output. 1-second resolution; LC_ALL=C for locale safety."""
+    try:
+        result = subprocess.run(
+            ["ps", "-p", str(pid), "-o", "lstart="],
+            capture_output=True, text=True, timeout=2.0, check=False,
+            env={**os.environ, "LC_ALL": "C"},
+        )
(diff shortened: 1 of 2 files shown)


junaidtitan:
Thank you for jumping on this, @ynaamane — proactively closing the residual I'd deferred in the #94 merge note is exactly the follow-through that's made your contributions so valuable. The engineering is solid: gate ordering is right (`_pid_is_alive` → `_pid_identity_match` → `_is_claude_process`), it's wired symmetrically into the `start_guard` watchdog, #94's mtime-immune liveness gate is preserved, and fail-OPEN is the correct posture (fail-closed would block legitimate reloads for everyone without psutil). `test_recycled_pid_no_resurrection` genuinely exercises the vector — I confirmed it fails with the gate neutered — and there's zero regression to the existing anti-resurrection suite.

Two things before we merge — one quick, one worth a quick design chat:

**1) (blocking, quick) Suite goes red on a default install.** The three `TestPidIdentityMatchRecycledPid` cases call the real `_record_claude_identity` → psutil, but cozempic ships **zero dependencies**, so psutil is absent on a default install and those three fail (we measured **3 failed / 859 passed** without psutil; the 862-green run had psutil incidentally present). A `pytest.importorskip(\"psutil\")` / `@unittest.skipUnless(...)` guard on those three fixes it.

**2) (design) The gate is a no-op without psutil — which is most installs.** psutil is an optional lazy import (never pulled in transitively), so `_get_pid_start_time` returns None on a default install and `_pid_identity_match` fails open → the recycling gate does nothing for the zero-dep majority. As-is it closes the vector only for users who independently have psutil.

Suggestion: source start-time from **stdlib** so it works for everyone, keeping your gate/identity structure exactly as-is — just swap where the timestamp comes from:
- **Linux:** `/proc/<pid>/stat` field 22 (`starttime`, clock ticks since boot) — pure file read.
- **macOS:** `ps -p <pid> -o lstart=` (or `-o etime=`) — we already shell to `ps` for identity, so it fits the existing pattern.
- Keep `psutil.create_time()` as a final fallback.

That makes the fix actually deliver the headline on a zero-dep install (and we'd soften the summary to note the prior psutil-only behavior).

Would you be up for taking the stdlib path here? Equally happy to layer it onto your branch ourselves if you'd rather — just say the word and we'll proceed either way. Genuinely grateful for closing this out.

ynaamane:
Thanks @junaidtitan — took the stdlib path. Two new commits on top of the original four:

- `bc1b078` — stdlib swap. `_get_pid_start_time` now dispatches: Linux `/proc/<pid>/stat` field 22 + `/proc/stat btime` (no subprocess, zero-dep); macOS `ps -p <pid> -o lstart=` with `LC_ALL=C` in subprocess env + `time.strptime` (zero-dep); Windows + final fallback `psutil.create_time()` (existing lazy-import). Same signature, same return type, same 0.1s tolerance, same fail-OPEN paths, same gate ordering.
- `4373ada` — round-2 adversarial review found one MED: `_get_pid_start_time_linux`'s `rfind(")")` returns `-1` on malformed `/proc` (fuse-/proc, WSL1, BSD-on-Linux emulation) → silent wrong slice. 2-line guard added.

**B1 + B2 empirically closed**: fresh `python -m venv` (pytest only, no psutil) → `pytest tests/test_guard_pid_recycling.py -v` → 17 passed, 5 Linux-skipped on macOS host. Independently re-verified in a separate fresh venv by our adversarial reviewer (841 passed full suite + 1 known pre-existing race flake unrelated to this PR).

Full suite: 869 passed, 5 skipped, 36 subtests — +7 tests (`TestLinuxStdlibStartTime` skipUnless Linux, `TestMacosStdlibStartTime`, `TestStdlibFallthrough`). PR body refreshed against the new commits per our internal pre-push checklist.

Softened the summary per your suggestion to note the prior psutil-only behavior. Process learning added internally — your framing ("does it not regress?" vs "does it deliver value?") is exactly the framing our 5-gate pipeline missed, now embedded as Rules 9-12 in our project's FIX DISCIPLINE.

Ready for re-review.

junaidtitan:
Thanks @ynaamane — this is the stdlib swap we needed. Confirmed the gate now actually works on a default zero-dep install (B1 + B2 both closed): it rejects a recycled live PID using `/proc/<pid>/stat` field-22 on Linux and `ps -p <pid> -o lstart=` on macOS, no psutil required. The paren-safe comm parse (rfind on ')') and the fail-open-to-1.8.16-behavior posture are right, and the recycled-PID integration test genuinely drives _terminate_and_resume with a fooled mtime + recycled start_time and asserts no SIGTERM. Merging.

PR #99, opened 2026-05-21
fix(doctor): show runnable treat commands for oversized sessions

carlkibler:
I have been running a Cozempic fork for a month and accumulated a few enhancements I like. This first one is cosmetic to give the exact commands a user can run to 'treat' a session. Just want to see if PRs are accepted, then will submit the others which have small improvements I found useful. 

## Problem

The `doctor` command suggested running the `treat` command, but didn't provide command the user can just copy & paste. 

## Change

Output now includes the exact commands to run:

```
Treat each oversized session:
  cozempic treat a1b2c3d4 -rx aggressive --execute
  cozempic treat e5f6a7b8 -rx aggressive --execute
```

No functional changes, just text output. 

## Tests

Four new unit tests in `tests/test_doctor.py` (using the same `patch("cozempic.doctor.find_sessions", ...)` pattern as neighboring tests):

- clean-session case still returns `status="ok"` with no fix_description
- fix_description contains real 8-char IDs, not `<session>`
- one `cozempic treat` line per offending session
- commands are sorted largest-first (same order as the message summary)

All 41 existing tests continue to pass.

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -855,7 +855,9 @@ def cmd_doctor(args):
         print(f"    {icon} {r.name:<25} [{r.status.upper()}]")
         print(f"      {r.message}")
         if r.fix_description and r.status not in ("ok", "fixed"):
-            print(f"      Fix: {r.fix_description}")
+            fix_prefix = "      Fix: "
+            fix_text = r.fix_description.replace("\n", f"\n{' ' * len(fix_prefix)}")
+            print(f"{fix_prefix}{fix_text}")
         print()
 
         if r.status == "issue":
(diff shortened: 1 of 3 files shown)


junaidtitan:
Merged — thank you, @carlkibler, and to answer your question directly: yes, PRs are very much welcome. This is exactly the kind of contribution to send — tight, well-scoped, and the "no functional change, just text" claim holds up under review.

We checked it across all three lenses: security (the session IDs come straight from `find_sessions()` / the JSONL filenames and are display-only — never touch a shell, no injection path), logic (the oversized-sessions pass/fail status is unchanged, the emitted `cozempic treat <id> -rx aggressive --execute` syntax is exactly right, and largest-first ordering matches the summary), and tests (866 green; your 4 new cases assert real 8-char IDs rather than the `<session>` placeholder, one line per session, and that the clean case still returns ok). Copy-pasteable commands are a real UX win for that check.

Please do send the rest of your queued enhancements — happy to review them. Genuinely glad to have you contributing.

PR #100, opened 2026-05-22
fix(_HostFileLock): normalize fd position to 0 before msvcrt.locking on Windows

GravyaDev:
Follow-up to #96 — applies the same `seek(0)` fix to `_HostFileLock.__enter__` (helpers.py:82).

`__exit__` already rewinds to byte 0 before `LK_UNLCK` (helpers.py:105). Without a matching `seek(0)` in `__enter__` before `LK_LOCK`, a stale non-empty lock file (left by a prior crashed run) causes the acquire and release calls to target different byte ranges and silently fail to serialize — same root cause as #96 fixed in `_SettingsLock`.

The fix is placed inside the `if os.name == "nt":` branch to match #96's pattern: Windows is the only platform where the seek has behavioral effect — `fcntl.flock` on POSIX is fd-based and position-independent.

The OSError cleanup path on `_HostFileLock.__enter__` is already correct (closes `self._fh` on degradation), so no parallel cleanup-symmetry work is needed here.

## Tests

`tests/test_host_file_lock.py` mirrors `tests/test_settings_lock.py`:

- `test_posix_acquires_fcntl_flock` — POSIX path baseline
- `test_windows_acquires_msvcrt_locking` — Windows acquire/release shape
- `test_windows_seek_zero_before_lock_and_unlock` — dual-seek symmetry
- `test_windows_acquire_position_normalized_on_stale_lock_file` — regression test that actually proves the bug is fixed (pre-populates the lock file with content, captures position at each `locking()` call, asserts both LK_LOCK and LK_UNLCK fire at byte 0)

Local run on Windows host: 3 passed, 1 skipped (POSIX test skips on Windows — same pattern as the POSIX tests in `test_settings_lock.py`). The PR #96 `test_settings_lock.py` suite remains green.

## Edge cases

| # | Scenario | Failure mode | Status |
|---|---|---|---|
| E1 | Fresh empty lock file | seek(0) is no-op | already-covered (pre-existing behavior) |
| E2 | Stale 1-byte lock file | LK_LOCK at byte 1, LK_UNLCK at byte 0 — leak | implemented (regression test) |
| E3 | Stale many-byte lock file | same as E2 with worse drift | implemented (regression test) |
| E4 | lock_path is a directory | IsADirectoryError | already-covered (`except OSError`) |
| E5 | lock_path.parent missing | FileNotFoundError | already-covered (mkdir on line 77) |
| E6 | Concurrent acquire | race | non-bug (LK_LOCK blocks; kernel serializes) |
| E7 | Partial-failure (open OK, lock fail) | leaked fd | already-covered (line 89-93 closes _fh) |
| E8 | seek(0) raises OSError | unlocked-but-thinks-locked | non-bug (degradation is the contract) |
| E9 | ImportError msvcrt | no lock available | already-covered |
| E10 | POSIX path | seek(0) skipped | non-bug (flock fd-based, position-independent) |
| E11 | Lock file deleted mid-hold | orthogonal | non-bug |
| E12 | Re-entrant enter | not supported by design | out-of-scope (same as _SettingsLock) |

---
*Disclaimer: this PR description, the code changes in 5598293, and the accompanying tests were drafted by @kloud-ai under human review and approval. Human in the loop: GravyaDev.*

### DIFF
--- a/src/cozempic/helpers.py
+++ b/src/cozempic/helpers.py
@@ -78,7 +78,16 @@ class _HostFileLock:
             self._fh = open(self._lock_path, "a")
             if os.name == "nt":
                 import msvcrt
-                # Lock first byte; blocking
+                # Windows — msvcrt.locking locks bytes from the CURRENT file
+                # position. "a" (append) mode leaves the pointer at EOF:
+                # byte 0 on a fresh empty lock file, but >0 if a stale
+                # non-empty lock file was left from a prior crashed run.
+                # __exit__ already rewinds to byte 0 before LK_UNLCK
+                # (helpers.py:105), so without this matching seek(0) before
+                # LK_LOCK the two operations would target different byte
+                # ranges and silently fail to serialize. Mirrors the
+                # _SettingsLock fix in init.py (PR #96).
+                self._fh.seek(0)
                 msvcrt.locking(self._fh.fileno(), msvcrt.LK_LOCK, 1)
             else:
                 import fcntl
--- a/tests/test_host_file_lock.py
+++ b/tests/test_host_file_lock.py
@@ -1,0 +1,196 @@
+"""Tests for _HostFileLock cross-platform behavior.
+
+Mirrors test_settings_lock.py — _HostFileLock uses the same
+msvcrt.locking (per-byte, position-relative) pattern on Windows,
+with the same stale-non-empty-lock-file class of bug.
+
+    POSIX  → fcntl.flock (fd-based, position-independent)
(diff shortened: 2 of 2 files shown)


junaidtitan:
Thanks @GravyaDev — clean follow-up to #96. The dual-seek symmetry and the stale-lock-file regression test (pre-populating the lock file with content, then asserting both LK_LOCK and LK_UNLCK fire at byte 0) is exactly the right way to prove it. Verified the POSIX path is untouched and there's no fd-leak in the degradation path. Merging — keep them coming.

PR #102, opened 2026-05-28
fix(safety): refuse to prune active sessions + floor + structural validation

ynaamane:
## Problem

The session pruner destroyed a live 63MB Claude Code session — truncating it to 730KB (98.8% loss) on its `compact-summary-collapse` pass. The pruned file became unresumable: `claude --resume <id>` hangs indefinitely with no error.

Forensic analysis of the corrupted session (`f8bf4ee2-8cee-4921-8c60-c77c68a6c66a`):

- Before: 11,978 entries, 9,881 unique UUIDs, **0 broken `parentUuid` pointers**.
- After: 612 entries, 213 unique UUIDs, **0 broken `parentUuid` pointers**.

So `_relink_parent_chain` in `executor.py` did its job — the damage isn't structural chain breakage. It's **semantic / volumetric**: 91% of assistants, 85% of users, 99% of `last-prompt` / `permission-mode` / `system` entries gone. Claude Code's resume engine needs the LAST `permission-mode`, the LAST `last-prompt`, the LAST `compact_boundary`, and a coherent conversation tail to bootstrap. The current pruner wipes them.

Root cause: `strategies/gentle.py::compact-summary-collapse` finds the last `compact_boundary` and removes everything before it. Its `_META_TYPES` exemption set was missing `permission-mode`, and there was no protection against pruning sessions that were still actively being written to. The guard daemon's 30s soft-prune cadence ran this strategy on a live session.

## Fix

Four orthogonal safety pillars:

### P0-A — Active-session idle guard
- New `src/cozempic/safety.py::assert_session_idle_or_force(path, *, min_idle_hours, force)` refuses to prune any session whose JSONL `mtime` is within `COZEMPIC_MIN_IDLE_HOURS` hours (default 24, configurable, clamped to `[0.0, 168.0]`).
- Plumbed through `guard.guard_prune_cycle` (logs + skips K-counter advance on refusal), `cli.cmd_treat`, `cli.cmd_reload` (exit code 4 with actionable message), `overflow.OverflowRecovery.on_file_growth` (auto-`force=True` for the 90% reactive emergency).
- New CLI flag `--force` on `treat` and `reload` for explicit operator override.

### P0-B — Post-prune structural validation
- New `safety.validate_post_prune(msgs_before, msgs_after)` runs checks C1–C6 BEFORE save:
  - C1: every surviving `parentUuid` resolves to a surviving UUID.
  - C2: the ORIGINAL session root (first `parentUuid=null` in `msgs_before`) survives — by UUID equality, not just "any null parent exists".
  - C3: at least one `user` and one `assistant` survive.
  - C4: if a `compact_boundary` existed, the LAST one survives.
  - C5: if `permission-mode` entries existed, the LAST one survives.
  - C6: if `last-prompt` entries existed, the LAST one survives.
- Failure → `PruneValidationError`, caller never calls `save_messages`. CLI exit code 5.
- Also runs INSIDE `session.save_messages` on the merged tmp file after the append-delta path, catching mid-prune writes from Claude that would otherwise bypass validation.

### P0-C — Floor preservation in executor
- New `safety.enforce_floor(msgs_before, msgs_after, *, cfg)` re-adds dropped must-preserve entries:
  - First `parentUuid=null` message (always — if it existed).
  - Last 50 `user` + 50 `assistant` messages (configurable).
  - Top up `user` + `assistant` survival to ≥ 50% (configurable, clamped to `[0.0, 1.0]`).
- Centralized at executor level (not per-strategy) to avoid "death by 10000 cuts" cumulative drops. Pipeline order: strategies → `enforce_floor` → `fix_orphaned_tool_results` → `validate_post_prune`. The floor-before-orphan-fix ordering prevents floor-rescued `tool_result` blocks from leaving orphans that would 400 the Anthropic API on resume.
- Honors `_relink_parent_chain` semantics — re-added entries get their `parentUuid` resolved forward through the effective-removals set.
- Floor enforcement is gated by an explicit `enable_floor: bool = True` kwarg on `run_prescription` rather than a config-dict flag (no hidden escape hatch in production code paths).

### P0-D — Last-of-type metadata singleton protection
- `executor._tag_last_of_metadata_types` pre-tags the LAST `ai-title`, `last-prompt`, `permission-mode` entry before strategies run. `helpers.is_protected` honors the tag. Tag is stripped before save so it never leaks to disk.
- Defense-in-depth: `permission-mode` also added to `strategies/gentle.py::_META_TYPES` set, so even without the singleton tag the strategy doesn't strip it.

## Tests

- 5 new test files + 1 fixture generator + 2 review-cycle test files = 8 new test surfaces, 46 new test methods, all GREEN.
  - `tests/test_session_prune_safety.py` (11 tests) — active-session guard
  - `tests/test_session_prune_validation.py` (21+ tests) — C1–C6 checks + executor wiring
  - `tests/test_session_prune_floor.py` (7 tests) — floor re-add + survival cap
  - `tests/test_session_prune_metadata.py` (6 tests) — singleton protection
  - `tests/test_session_prune_e2e.py` (1 regression test) — end-to-end with synthetic ~1MB fixture mimicking the corrupted session shape
  - `tests/test_session_prune_orphans.py` (1 test) — floor + orphan-fix ordering
  - `tests/test_session_save_append_validation.py` (1 test) — H-3 mid-prune append window
  - `tests/_session_factory.py` + `tests/conftest.py` — deterministic fixture generator
- Updated 2 pre-existing tests to reflect the new floor behavior:
  - `tests/test_guard_futile_reload.py::_make_session_file` — aged mtime so the new active-session guard doesn't refuse the test fixtures.
  - `tests/test_track2.py::test_gentle_with_boundary` — floor now preserves the first 2 messages on the 4-message fixture.
- Full suite: **875 passed, 36 subtests passed, 0 failures, 7 skipped, 34.22s** — `PYTHONPATH=src python3 -m pytest tests/ -q --ignore=tests/test_model_detection.py` (the 3 ignored `test_model_detection` failures are pre-existing on `main`).

## Backward compatibility

This is a behavior change: the guard daemon + `cozempic treat` / `reload` now refuse by default to prune sessions whose JSONL was modified in the last 24 hours.

- Operators who explicitly want the old always-prune behavior: set `COZEMPIC_MIN_IDLE_HOURS=0` or pass `--force` on the CLI.
- New exit codes: 4 (active session — use `--force`), 5 (post-prune validation failure). Documented in `--help`.
- Reactive overflow recovery (`overflow.OverflowRecovery.on_file_growth`) is exempt — it passes `force=True` because the 90% emergency case IS the active-session case.
- Zero new dependencies. cozempic stays zero-dep — both new modules (`safety.py`, `config.py`) use stdlib only.

## Process trail

Internal team-lead process artifacts (`PLAN.md` in the repo, `REVIEWER_PROBES.md` + `REVIEW_REPORT.md` kept local) followed the BMAD-style audit/RED/GREEN/adversarial-review pipeline established on PRs #84/85/86/88/89/92. The review pass surfaced one CRITICAL finding (floor + orphan-fix ordering) and three HIGH findings — all four fixed in this PR (commits `6f5b2a3`, `549fb19`, `9eeebd0`, `f76588b`).

## Scope

In scope (this PR):
- Safety guard + floor + validation + metadata singleton protection
- CLI plumbing (`--force` flag, exit codes 4/5)
- Documentation: README + strategy decorator metadata + commit-level CHANGELOG entries

Out of scope (deferred):
- Auto-restore UX (prompt user when resume fails) — P2 in the bug handoff
- `cozempic log` command listing recent prune outcomes — P2
- ECC plugin Rust error wrapping (`manager.rs:1544`) — different repository


junaidtitan (changes requested):
Thanks @ynaamane — this is a lot of careful work and the **instinct is exactly right**: floor preservation (P0-C), structural validation (P0-B), and last-of-type metadata protection (P0-D) are all genuinely good, and I verified end-to-end that they *do* prevent the 63MB→730KB class of wipe (floor re-adds root + ≥50% of users/assistants, forward-relinks the chain, no dangling parents; the `permission-mode` addition to `gentle._META_TYPES` + the singleton tag both hold). The C2 "original root survives by UUID equality" logic and the floor→orphan-fix→validate ordering are correct. Keep all of that.

But I can't merge it yet — a 3-lens review (security / logic / test, on copies of real sessions) surfaced one **CRITICAL** functional regression, two **HIGH** issues, and a couple of **MED** ungated paths. Two of these I re-verified by hand in the diff.

### 🔴 CRITICAL — the guard can no longer prune (its core function)
All four daemon tier call sites — `guard.py:606` (HARD2/80%), `:641` & `:653` (HARD/55%), `:850` (SOFT/25%) — call `guard_prune_cycle(...)` with the default `force=False`. The new idle gate (`guard.py:941`, post-lock recheck `:964` → `assert_session_idle_or_force`) then refuses any session whose mtime is within `COZEMPIC_MIN_IDLE_HOURS` (default 24h). **A session that crosses a token threshold is by definition being actively written (mtime ≈ now), so every threshold prune is refused.** The daemon's 25/55/80% auto-prune+reload never fires on a real session. `tests/test_session_prune_safety.py::test_guard_prune_cycle_skips_active_session_and_returns_no_change` actually codifies this refusal as expected — so the suite stays green while the feature is off. The idle guard should protect *user-initiated* `treat`/`reload` against someone else's live session, not block the daemon's own managed reload.

### 🟠 HIGH — #106's data-loss race is still open on the overflow path
`overflow.py:217` runs `guard_prune_cycle(force=True)` (in-place `os.replace` via `session.py`) **before** `_terminate_and_resume` at `overflow.py:273`. So at the 90% emergency — the highest-risk moment — we still `os.replace` a live inode under Claude's open fd. Both #106 gaps (the classify→replace TOCTOU and the inode-swap) remain reachable exactly there. The append-delta re-validation is good mitigation but, as #106 notes, can't fully close it while a live process holds the fd.

### 🟠 HIGH — C1 false-positive bricks resumed/forked sessions
`validate_post_prune` C1 (`safety.py:354-370`) fails on any surviving `parentUuid` not in the surviving set, with **no `msgs_before` baseline** (unlike C2). A resumed session whose first in-file message points at an out-of-file parent fails C1 even on a *zero-removal* prune → exit 5, unprunable even with `gentle`. Reproduced deterministically on a real session. `simulate_replay_readiness` (`safety.py:395-403`) has the same flaw. Fix: only fail C1 for parents that resolved in `msgs_before` but dangle in `msgs_after` (i.e. breakage the prune introduced) — mirror C2's baseline-relative check.

### 🟡 MED — two ungated write paths
`cmd_strategy --execute` (`cli.py:462`) and `doctor.fix_orphaned_tool_results` (`doctor.py:744`) both `os.replace` potentially-live sessions with no idle gate and no `--force`. The PR plumbs P0-A into treat/reload/guard but misses these two.

### Recommended direction
The clean fix for both the CRITICAL and the HIGH-overflow is the same one #106 recommends: **terminate Claude first, then prune the now-quiesced file, then resume** — for the daemon's managed reload tiers *and* the overflow path. Then the in-place write happens on a file no one holds open, `force=True` is safe there, and the idle guard can stay strict for everything else. Keep P0-B/C/D as-is (just make C1 baseline-relative), and add the idle gate + `--force` to `cmd_strategy`/doctor.

Also: please drop `PLAN.md` from the repo root before the next push (keep the process docs local, like the REVIEWER_PROBES/REVIEW_REPORT ones). And a real regression test for the #106 race itself (hold a live writer fd, append a sentinel after classify but before replace, assert refuse-or-preserve) would lock this down — @Snailflyer sketched the exact shape on #106.

Happy to pair on the terminate-first restructure if useful — it touches the anti-resurrection logic so it's worth doing together. Really solid foundation here; it's the daemon-reload sequencing that needs the rethink.

ynaamane:
Thanks @junaidtitan — this was the right call, and the terminate-first direction was exactly the fix. Pushed the restructure; here's what changed and the one item I want your read on.

## 🔴 CRITICAL — daemon can prune again (terminate-first, not just `force=True`)
You were right that the idle gate blocked every daemon tier. But a bare `force=True` would have been worse than the disease: `guard_prune_cycle` already did `save_messages` (in-place `os.replace`, session.py) **before** `_terminate_and_resume` — so turning the tiers on would have fired the #106 live-inode race on every reload. The real fix is the terminate-first restructure you suggested, and it required **splitting** `_terminate_and_resume`:

- `_terminate_claude(pid, …) -> TERMINATED | ALREADY_GONE | SKIPPED_SSH | FAILED_TO_DIE` — anti-resurrection entry gate (`_pid_is_alive` + `_pid_identity_match` + `_is_claude_process`) + kill + wait. No resume, no sentinel.
- `_resume_claude(…)` — resume only (tmux/screen send-keys, or `_spawn_reload_watcher`), sentinel written here.
- `_terminate_and_resume` is kept as their composition, so existing callers/tests are unchanged.

Why the split: `_terminate_and_resume` *resumes* (tmux/screen send `claude --resume` in-pane; plain spawns a watcher that resumes on death). Calling it then pruning would reopen the file under a fresh Claude before `os.replace` — the race just moves. So the daemon reload path is now: `_ReloadLock` → `_PruneLock` → snapshot+load+prune (read-only) → **futility check BEFORE terminate** (we never kill Claude for a futile prune) → `_terminate_claude` → `save_messages` on the **quiesced** file → `_resume_claude`. `FAILED_TO_DIE` aborts (no `os.replace` on a live file); `ALREADY_GONE` saves-without-resume (anti-resurrection); lock-order invariant `_ReloadLock ⊃ _PruneLock` documented. `force=True` on all 4 daemon tiers; the idle gate stays strict (`force=False`) for user-initiated `treat`/`reload`/`strategy` + doctor.

I rewrote the test that codified the refusal (`test_guard_prune_cycle_skips_active_session…`) to assert the daemon now prunes with `force=True`, and added a complementary test that `force=False` **still** refuses (P0-A intact for user paths).

**This is the anti-resurrection path you offered to pair on — I'd genuinely welcome your eyes on the split** (`_terminate_claude`/`_resume_claude` + the daemon/overflow sequencing). Two residual risks an internal review surfaced and fixed: (R-1) `terminate_failed` added to the K-counter skip-list so a failed kill doesn't march the circuit breaker; (R-2) `_detect_claude_flags` needs a live PID, so original CLI flags are now captured *before* the kill and threaded into resume (incl. `_spawn_reload_watcher`).

## 🟠 HIGH — overflow terminate-first
`_do_recover` now terminates first, then prunes the quiesced file, then resumes — same split. It aborts the prune on `FAILED_TO_DIE` **and** `SKIPPED_SSH` (an internal pass caught that an SSH session leaves Claude alive — pruning there would have reintroduced exactly the #106 race; the daemon path already returned `_no_change`, overflow now matches).

## 🟠 HIGH — C1 false-positive (incl. your zero-removal case)
C1 is now unconditionally baseline-relative, mirroring C2: a `parentUuid` absent from `msgs_before` is a cross-session anchor and is skipped; C1 only raises when the parent existed pre-prune but the prune removed it. Verified the exact symptom you named — a **zero-removal** prune of a resumed session (`validate_post_prune(msgs, list(msgs))`, external root parent) no longer exits 5 — with a dedicated regression test, and the H-3 append-orphan case still fires correctly (real pre-prune messages are threaded as `messages_before_prune`).

## 🟡 MED — ungated write paths
Idle gate + `--force` added to `cmd_strategy --execute`; idle gate (force=False, skip-active) added to `doctor.fix_orphaned_tool_results`.

## `simulate_replay_readiness` — fixed (your C1-parity finding)
Fixed behaviorally, not just documented: it now treats a `parentUuid` absent from the file as a cross-session anchor (resumed/forked head) instead of a chain break, and the old "no root" check is replaced by a `has_anchor` check (a pure resumed session has no `parentUuid=null` entry — that was the same false positive). The docstring is honest about what a single-list probe genuinely *can't* see (a removed root or a pruned mid-chain parent are indistinguishable from a cross-session anchor without a before/after baseline) — that detection lives in the two-list `validate_post_prune`. It's test-only today (zero `src/` callers), but it's in `__all__`, so the false positive is gone for any future caller too.

## Kept as-is per your review
P0-B (structural validation), P0-C (floor), P0-D (last-of-type metadata) untouched — only C1 changed. `PLAN.md` dropped from the repo root.

I merged `main` into the branch (rather than a 20-commit rebase over the anti-resurrection path) to absorb #98/#103/#104/#107 — say the word if you'd prefer it rebased. Full suite green: **1070 passed, 5 skipped** (`pytest tests/ --ignore=tests/test_model_detection.py`). Fix commits on top of the `main` merge: `b618583` (RED tests) · `67d99a9` `5f08fea` (C1 baseline-relative) · `63bddd8` (cmd_strategy/doctor idle gate) · `bf404ad` (terminate-first split + overflow) · `4e568fa` `1da4a14` (simulate_replay) · `25f0a75` (drop PLAN.md).

junaidtitan:
Update + thank-you, @ynaamane. **#106 is fixed and shipped in 1.8.18.**

Your work here drove it: the forensic characterization (the 63MB→730KB trace) nailed *what* was breaking, and working through this PR is exactly what surfaced that the real fix is **terminate-first** rather than an idle-guard. 1.8.18 implements that — the guard now never `os.replace`s a live transcript (read-only on the continuous tiers; terminate → prune → resume on reload/overflow, with an append-aware post-death write). The live-fd regression test shape is in the suite too.

I'm closing this PR because its headline (#106) is resolved and the idle-guard approach here would disable the daemon's own pruning (the blocker from the review). **But P0-B/C/D are genuinely valuable and I don't want to lose them** — the post-prune structural validation (C1–C6, with the C1 fix to be `msgs_before`-relative), the floor preservation, and the last-of-type metadata protection are great *defense-in-depth* that's orthogonal to the write-timing fix. If you're up for it, a focused PR with just those three (no idle-guard / no terminate changes — 1.8.18 owns those now) would be a strong follow-up, and I'll fast-track the review. Genuinely appreciate the depth you brought here.

PR #103, opened 2026-05-28
fix(cli,digest): polish v3 — digest --session UUID resolution + COZEMPIC_DEBUG token parsing + ValueError dispatch coverage

ynaamane:
## Problem

Three small, long-parked defects (queued since the polish-v2 audit) plus one CLI bug surfaced by live smoke testing:

1. **`cozempic digest --session <UUID>` crashes.** `_digest_session` returned the raw `--session` argument as a file path, so passing a session ID or UUID prefix — which the `--session` help advertises as supported — raised `FileNotFoundError` deep in `load_messages`. Only a full `.jsonl` path worked.
2. **`COZEMPIC_DEBUG` only honoured the literal `"1"`.** `COZEMPIC_DEBUG=true` / `yes` / `on` were silently ignored — inconsistent with common env-var conventions and surprising for a debug toggle.
3. **The `main()` `except ValueError` dispatch had zero test coverage.** The `guard`/`reload` error path (exit 2 + clean one-line stderr) versus the re-raise path for every other command was untested, so a refactor could silently break the user-facing error message.

## Fix

- **`parse_env_bool` helper** (new, `_validation.py`) — sits beside the existing `parse_env_positive_int` / `parse_env_non_negative_int` and follows the same warn-and-fallback contract. Accepts `1/true/yes/on` (truthy) and `0/false/no/off` (falsy), case-insensitive and whitespace-stripped; an unrecognized value warns once and falls back to the default. A `warn` parameter lets hot re-read paths suppress per-call spam.
- **`digest.py`** consumes `parse_env_bool` at both the module-level cache and the in-body `_debug()` re-read (module-load warns once; the in-body re-read uses `warn=False`) — a single parse path the two sites can no longer diverge on.
- **`_digest_session`** now routes an explicit `--session <UUID | prefix | path>` through the existing `resolve_session()` and recovers `session_id` from the resolved path stem. Absent or the `current` keyword take the same cwd-based `find_current_session(cwd)` auto-detect the no-arg path already used (kept consistent on purpose, so `--session current` does not silently switch to a different detection strategy).
- **ValueError dispatch coverage** adds `TestValueErrorDispatch` exercising both branches (test-only — the production code was already correct).

F7 from the original polish-v2 list (`COZEMPIC_DEBUG` captured at import time) was already fixed on `main` — `_debug()` re-reads the env on each call. Verified; no change needed.

## Tests

- **866 passing** (`pytest tests/ -q --ignore=tests/test_model_detection.py`, the standing exclusion), exit 0.
- Adds four test classes — `TestParseEnvBool`, `TestDebugFlagTokens`, `TestValueErrorDispatch`, `TestDigestSessionResolution` — covering token broadening (incl. case-insensitive, whitespace, explicit-falsy, warn suppression, whitespace-only silent), import-time reload, both dispatch branches, and UUID/prefix/path/`current`/no-arg resolution.
- Strict RED→GREEN: each fix has a test verified to fail at the pre-fix state (confirmed by reverting each fix and re-running).
- Zero new dependencies — pure stdlib (`os`/`sys`/`typing` + intra-package).

## Scope

Touches only `_validation.py`, `digest.py`, `cli.py` and their tests — no `guard.py` / `executor.py`.

**Deferred (documented, not folded):** the `COZEMPIC_NO_TELEMETRY` / `NO_AUTO_UPDATE` / `NO_GLOBAL_INIT` / `NO_AUTO_INIT` opt-out flags still use bare truthiness (so `=false` would still *enable* the opt-out). Migrating them to `parse_env_bool` is a behaviour change with back-compat implications for those opt-out semantics, so it is left as a follow-up rather than bundled here.

**Pre-existing, tracked follow-up:** `digest update` / `flush` resolve the current session non-strict (mtime fallback can guess on ambiguity), unlike `treat` which passes `strict=execute`. This predates the PR (base `3a4cb1c` is also non-strict) and `flush` runs in the Stop hook where `strict=True` could fail silently — it needs its own analysis, not a blind fold here.

**Known behaviour:** the module-load `parse_env_bool` now warns once on a genuinely-invalid `COZEMPIC_DEBUG` value (the old `== "1"` was silent). Hook paths suppress stderr via `2>/dev/null`, so this surfaces only in interactive use and is actionable. The import-time reload test relies on test-collection order-independence (verified).

### DIFF
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -206,3 +206,40 @@ def parse_env_non_negative_int(name: str) -> int | None:
         _env_warn(name, raw, "must be non-negative")
         return None
     return value
+
+
+# Token sets are module-level constants so tests can import and inspect them.
+_BOOL_TRUE_TOKENS: frozenset[str] = frozenset({"1", "true", "yes", "on"})
+_BOOL_FALSE_TOKENS: frozenset[str] = frozenset({"0", "false", "no", "off"})
+
+
+def parse_env_bool(name: str, default: bool = False, warn: bool = True) -> bool:
+    """Read env var `name`, parse as boolean, return `default` if absent.
+
+    Truthy tokens (case-insensitive, whitespace stripped): 1, true, yes, on
+    Falsy tokens  (case-insensitive, whitespace stripped): 0, false, no, off
+    Unrecognized non-empty value: if `warn`, emits a warning to stderr and
+    returns `default`.  Empty / absent: returns `default` silently.
+
+    Follows the warn+fallback contract of parse_env_positive_int:
+    env vars are ambient config; an unrecognized value must not crash.
+
+    The `warn` parameter suppresses the warning when set to False — useful
+    for in-body re-reads that run on every call (e.g. _debug()) where a
+    single module-load warning is enough and per-call spam would drown output.
+    """
+    raw = os.environ.get(name)
+    if raw is None or raw == "":
+        return default
+    normalized = raw.strip().lower()
+    if normalized in _BOOL_TRUE_TOKENS:
+        return True
+    if normalized in _BOOL_FALSE_TOKENS:
+        return False
+    if warn:
+        _env_warn(
+            name,
+            raw,
+            f"must be one of {sorted(_BOOL_TRUE_TOKENS | _BOOL_FALSE_TOKENS)}",
+        )
+    return default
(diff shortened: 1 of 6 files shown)


junaidtitan:
Thanks @ynaamane — solid polish-v3. `parse_env_bool` slotting in beside `parse_env_positive_int` with the warn/warn=False split for the hot re-read is clean, and routing `digest --session` through `resolve_session` (recovering the full UUID from the stem) fixes the crash properly. Verified resolver parity with the other subcommands and that the ValueError-dispatch tests assert the real exit-2 path. Merging.

PR #104, opened 2026-05-28
fix(recap): honour max_turns + symlink-safe write + injection sanitization + ReDoS cap (recap.py audit)

ynaamane:
## Problem

`recap.py` generates the "PREVIOUSLY ON THIS SESSION" summary that is `cat`'d into the new terminal before `claude --resume` (cli.py). It had **zero test coverage** and several latent defects:

1. **Dead `max_turns` parameter.** `generate_recap(messages, max_turns=40)` (and `save_recap`) accept `max_turns` but the body iterates *all* messages — the parameter has no effect. A large session has every turn processed for topic analysis.
2. **`save_recap` follows symlinks.** It used `dest.write_text(...)`, which follows a symlink at the destination. Every other write path in the package uses the crash-safe, symlink-safe `atomic_write_text` helper.
3. **Prompt-injection vector via the recap.** The recap is read back into the *next* session's context (it is `cat`'d into the resuming terminal). User turns were sanitized with `_clean_user_text`, but the assistant turn shown on the `Last:` line was only whitespace-collapsed — so a compromised/adversarial assistant turn containing e.g. `<system-reminder>…</system-reminder>` survived verbatim into the next session.
4. **`_truncate` negative-index slice.** For `max_len < 3`, `text[:max_len-3]` is a reverse slice and returns output *longer* than `max_len`.
5. **`O(n²)` tag-regex blow-up (found in review).** `_clean_user_text`'s generic-tag regexes (`<[^>]+>.*?</[^>]+>`, `<[^>]+/?>`) are quadratic. On realistic `<`-heavy content they block the reload path: git-conflict markers (`<<<<<<< HEAD` ×3000) take ~2.3s; pathological input (~28s). Fix #3 widened the exposure because assistant text (long, code/JSX/C++/conflict-heavy) now flows through these regexes too.
6. Theme labels could carry a trailing `_`/`-` (`"unique_ (3)"`).

## Fix

- **`max_turns` is honoured** — collect all user turns and keep the full-session `total_exchanges` count for the header, but window topic/theme analysis to the most-recent `max_turns` turns.
- **`save_recap` uses `atomic_write_text`** — symlink-safe + crash-safe, consistent with the rest of the package (verified: a pre-planted symlink at the target is replaced by a regular file, the victim file is untouched).
- **`_clean_user_text` is applied to the assistant turn** as well as user turns, stripping system/command tags before they can re-enter the next session's context. (Note: tag-shaped content is removed; plain-text instructions are indistinguishable from legitimate text and are not — see Scope.)
- **Input length-cap in `_clean_user_text`** bounds the quadratic tag regexes (the recap only ever displays ~72-char snippets, so the cap loses no displayed information): pathological `<`×100000 drops from ~28s to ~181ms. This also fixes the pre-existing quadratic cost on the user path.
- **`_truncate` guard** returns `text[:max(max_len, 0)]` for `max_len <= 3`, so output never exceeds `max_len`.
- **Theme labels** strip trailing `_`/`-`.

## Tests

- New `tests/test_recap.py` (was zero coverage) — covers truncate boundaries, atomic-write, `max_turns` windowing vs full exchange count, injection sanitization of the assistant line, theme-label hygiene, the input length-cap, the `max_turns <= 0` / negative-value guard, and the happy path.
- Strict RED→GREEN: the bug-targeting tests fail at the pre-fix state.
- Also guards `_truncate` and the `max_turns` window against degenerate inputs: `max_turns=0` no longer aliases to the full list via `list[-0:]`, and a negative `max_turns` no longer drops the first turn via `list[1:]` (both now yield an empty topic window; `max_turns` is a non-negative limit).
- **860 passing** (`pytest tests/ -q --ignore=tests/test_model_detection.py`), exit 0. The `test_spawn_lock` / `test_overflow` race-stress tests are intermittent flakes pre-existing at the clean base `3a4cb1c` (verified in isolation; their modules are not in this diff).
- Zero new dependencies — pure stdlib.

## Scope

Touches only `recap.py` and `tests/test_recap.py`.

**Deferred (documented, not folded):** the recap output is `cat`'d into the resuming terminal verbatim, so plain-text (non-tag) adversarial instructions in a turn still survive `_clean_user_text`. Fully neutralizing that requires either fencing the recap or changing how the caller injects it — a `cli.py` change, kept out of this PR. The tag-stripping here removes the structured-tag vector; the residual is tracked as a follow-up.

**Accepted trade-off:** stripping tags from the assistant turn can remove legitimate angle-bracket code from the 72-char `Last:` preview — preferred over leaving the injection vector open.

**Performance:** `_clean_user_text`'s two *generic* tag regexes are O(n²). The named-tag regexes run first on the full text (so a system/command tag whose close lies past the cap is still removed), then a length-cap (input → 8000 chars; the recap only displays ~72-char snippets so nothing displayed is lost) bounds the generic regexes. Realistic `<`-heavy content (git-conflict markers, C++/JSX) drops from ~28s of stall on the resume path to ~180ms. A degenerate all-openers flood (>100 KB of literal `<system-reminder>` repeats in a single turn) is ~1.4s on the named-tag pass — perf-only, not an injection, and not produced by real turns.

### DIFF
--- a/src/cozempic/recap.py
+++ b/src/cozempic/recap.py
@@ -42,16 +42,20 @@ def _extract_text(msg: dict) -> str:
 
 def _clean_user_text(text: str) -> str:
     """Remove system tags, command noise, and whitespace from user text."""
-    # Cap input to bound the O(n²) tag regexes below. A git-conflict-heavy
-    # assistant turn (e.g. `<<<<<<< HEAD` × 3000) otherwise blocks the
-    # reload path for seconds. 8000 chars >> any displayed snippet (~72 chars).
-    text = text[:8000]
+    # Strip NAMED system/command tags on the FULL text first so that an injection
+    # tag whose close-tag lies past the cap below is still removed (N-1).
+    # These regexes are anchored on literal tag names → linear on realistic input.
+    # LOW residual: a pathological <named-tag>×N flood would be O(n²) here too,
+    # but that is a perf-only concern (not an injection vector) and absurd in practice.
     text = re.sub(r"<system-reminder>.*?</system-reminder>", "", text, flags=re.DOTALL)
     text = re.sub(r"<local-command-caveat>.*?</local-command-caveat>", "", text, flags=re.DOTALL)
     text = re.sub(r"<command-name>.*?</command-name>", "", text, flags=re.DOTALL)
     text = re.sub(r"<command-message>.*?</command-message>", "", text, flags=re.DOTALL)
     text = re.sub(r"<command-args>.*?</command-args>", "", text, flags=re.DOTALL)
     text = re.sub(r"<local-command-stdout>.*?</local-command-stdout>", "", text, flags=re.DOTALL)
+    # Cap before the GENERIC tag regexes (the true O(n²) ones). Recap displays
+    # ~72-char snippets so the cap loses no displayed info.
+    text = text[:8000]
     text = re.sub(r"<[^>]+>.*?</[^>]+>", "", text, flags=re.DOTALL)
     text = re.sub(r"<[^>]+/?>", "", text)
     text = re.sub(r"SessionStart:.*", "", text)
--- a/src/cozempic/recap.py
+++ b/src/cozempic/recap.py
@@ -65,9 +65,11 @@ def _clean_user_text(text: str) -> str:
 
 def _truncate(text: str, max_len: int = 70) -> str:
     """Truncate text to max_len, adding ellipsis if needed."""
-    if len(text) > max_len:
-        return text[: max_len - 3] + "..."
-    return text
+    if len(text) <= max_len:
+        return text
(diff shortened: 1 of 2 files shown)


junaidtitan:
Thanks @ynaamane — the injection fix is the important one and it's done right: the assistant `Last:` line now flows through the full `_clean_user_text` (not just whitespace-collapse), and I couldn't find a control-token bypass (nested/attr/case/newline/split-block all neutralized). max_turns, the symlink-safe atomic_write swap, and the _truncate negative-index fix all check out. Merging. One follow-up I'm landing on main: the named-tag regexes still run on *uncapped* input, so a pathological assistant turn with thousands of literal `<system-reminder>` substrings can hit O(n²) (~20s on the reload path). I'll cap input length before those regexes — flagging so you have the context.

PR #107, opened 2026-05-29
fix(watcher): detect growth after prune (shrink + inode-swap), missing-file fallback, deterministic kqueue close (watcher.py audit)

ynaamane:
## Problem

`watcher.py`'s `JsonlWatcher` powers the guard's reactive overflow recovery — it watches the session JSONL for growth (kqueue on macOS, `os.stat` polling elsewhere) and fires `OverflowRecovery.on_file_growth`. The module had no dedicated test file and several defects, two of which silently disable overflow detection after a normal prune:

1. **Blind after a prune (shrink).** Both watch paths fire only when `new_size > self._last_size`. A guard prune shrinks the JSONL (e.g. 50 MB → 5 MB), but `_last_size` keeps the old 50 MB, so growth is not reported again until the file re-exceeds its pre-prune size — exactly when overflow recovery should stay armed.
2. **Blind after a prune (kqueue inode swap).** Prunes write via `save_messages` → `os.replace`, which swaps the file's inode. The kqueue path watches the *old* fd's inode and never sees the new file's growth, and it did not register `KQ_NOTE_DELETE`/`KQ_NOTE_RENAME` to notice the swap. Independent from #1 — even with the shrink fix, the kqueue path stayed dead post-prune.
3. **Crash when the session file doesn't exist yet.** The kqueue path called `os.open(filepath, O_RDONLY)` unguarded → `FileNotFoundError` killed the watcher thread (the poll path degrades gracefully).
4. **Silent callback errors.** Both paths did `except Exception: pass`, so a persistently-failing `on_growth` callback was invisible.
5. **kqueue fd not closed deterministically.** The kqueue object was never `.close()`d; on CPython the fd is reclaimed by GC at function return, but relying on GC is fragile (PyPy / reference cycles).
6. `__import__("select")` idiom re-imported `select` on every instantiation.

## Fix

All changes are internal to `watcher.py` — `JsonlWatcher.__init__` / `start` / `stop` signatures are unchanged (the `guard.py` caller is untouched).

- **Shrink reset (both paths):** when `new_size < self._last_size`, reset the baseline so growth is detected from the pruned size. The overflow callback's own danger-threshold fast-path absorbs any sub-threshold calls.
- **kqueue inode-swap → poll fallback:** register `KQ_NOTE_DELETE | KQ_NOTE_RENAME`; on such an event, close the kqueue + fd and fall back to the polling watcher, which re-stats *by path* and is therefore immune to inode swaps. (Trade-off: after the first prune the watcher polls at 200 ms instead of kqueue's sub-ms wake — negligible for overflow detection, whose surrounding loop runs on a 30 s cadence.)
- **kqueue path never crashes the daemon thread — degrades to poll on ANY `OSError`:** `os.open` catches `OSError` (not just `FileNotFoundError`, so `PermissionError`/EACCES also fall back); `select.kqueue()` and the `kq.control()` loop are wrapped so an `OSError` (e.g. `EMFILE` under fd exhaustion) sets a fallback flag and degrades to polling instead of letting the exception kill the watcher thread silently (which would stop overflow recovery for the session).
- **Callback errors → one-shot stderr log, reset on success:** the first `on_growth` exception is logged (`[watcher] on_growth error: …`); repeats are suppressed via a private flag (reset after any successful callback, so a recovery re-arms logging) — a persistently-failing callback can't flood stderr.
- **Deterministic, leak-proof close:** `kq.close()` and `os.close(fd)` run in `finally` (no longer relying on GC), and cleanup-close `OSError`s are swallowed so a failing `kq.close()` can neither leak the fd nor skip the poll fallback.
- **Import hygiene:** `_HAS_KQUEUE` resolved once at module load.

## Tests

- New `tests/test_watcher.py` — covers shrink-then-growth (both paths), missing-file fallback, kqueue inode-replace → poll takeover, one-shot callback-error logging, and the kqueue-close/import-constant behaviors. kqueue-specific cases are `@skipUnless(sys.platform == "darwin")`; all watcher-thread tests use `daemon=True` threads with bounded `join` timeouts so the suite never hangs. (The existing poll-mode `TestJsonlWatcher` in `test_overflow.py` is left as-is.)
- **845 passing** (`pytest tests/ -q --ignore=tests/test_model_detection.py`), exit 0 — `tests/test_watcher.py` adds 20 tests, including the OSError-robustness set (PermissionError→poll, `kqueue()` EMFILE→poll, fd closed even when `kq.close()` raises, kqueue thread survives cleanup errors and still polls). The behavior-bug tests are genuinely RED at the pre-fix state — verified via broken-variant shims, not tautological. A few `test_spawn_lock` / `test_overflow` / `test_guard_hardening` race-stress tests are intermittent flakes pre-existing at the clean base `3a4cb1c` under heavy parallel load (their modules are not in this diff; they pass in isolation and sequentially).
- Zero new dependencies — pure stdlib.

## Scope

Touches only `watcher.py` and `tests/test_watcher.py`. No API change, so the `guard.py` / `overflow.py` call sites are untouched. The behavioral changes (shrink reset, inode-swap fallback) make the watcher report *more* of the growth it should already have reported — existing callers see correct callbacks, never fewer.

### DIFF
--- a/src/cozempic/watcher.py
+++ b/src/cozempic/watcher.py
@@ -96,6 +96,7 @@ class JsonlWatcher:
                         self._last_size = new_size
                         try:
                             self.on_growth(self.filepath, new_size)
+                            self._cb_error_logged = False  # reset on success
                         except Exception as exc:
                             self._log_cb_error(exc)  # don't crash the watcher thread
                 if inode_replaced:
@@ -119,5 +120,6 @@ class JsonlWatcher:
                 self._last_size = new_size
                 try:
                     self.on_growth(self.filepath, new_size)
+                    self._cb_error_logged = False  # reset on success
                 except Exception as exc:
                     self._log_cb_error(exc)  # don't crash the watcher thread
--- a/src/cozempic/watcher.py
+++ b/src/cozempic/watcher.py
@@ -57,19 +57,24 @@ class JsonlWatcher:
     def _watch_kqueue(self) -> None:
         """macOS kqueue watcher — 0.04ms wake latency, 0% CPU idle.
 
-        Falls back to poll if the watched inode is replaced (os.replace / atomic prune).
+        Falls back to poll on:
+        - OSError from os.open (file missing, permission denied, etc.)
+        - OSError from select.kqueue() or kq.control() (EMFILE, etc.)
+        - Inode replacement detected via KQ_NOTE_DELETE/RENAME (os.replace / atomic prune)
         """
         import select
 
         try:
             fd = os.open(self.filepath, os.O_RDONLY)
-        except FileNotFoundError:
-            # File not yet created — poll path handles OSError gracefully
+        except OSError:
+            # File not yet created, permission denied, or other OS error —
+            # poll path handles OSError in _get_size() gracefully
             self._watch_poll()
             return
-        # kq = None sentinel: if select.kqueue() raises (e.g. EMFILE), the finally
-        # block below must still close fd — deterministic, not GC-dependent.
+        # kq = None sentinel: ensures the finally block can guard kq.close()
+        # independently of whether kqueue() succeeded.
         kq = None
+        kqueue_error = False
         inode_replaced = False
         try:
             kq = select.kqueue()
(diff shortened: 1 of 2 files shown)


junaidtitan:
Thanks @ynaamane — good catch on the post-prune blindness (both the shrink baseline staying stale *and* the kqueue inode-swap after os.replace). Verified this is safe to land independent of the #102/#106 work: the overflow path already snapshots the session and does append-aware classify, so this makes detection *correct* without widening the write-race — it changes *when* a prune fires, not *how* it writes. Live kqueue tests exercising the real inode swap are a nice touch. Merging.

PR #108, opened 2026-05-30
fix(session): normalize project slug (_,. → -) and use strict cross-project resolution

ynaamane:
## Problem

After a `/compact`, the **PostCompact** hook could inject a **different project's** `team-checkpoint.md` into the recovered context. A session in project `foo_bar` (underscore in the dir name) would recover another, unrelated project's team state if that other project's session was more recently active at compaction time.

Root cause is two stacked bugs:

1. **Slug normalization mismatch.** `cwd_to_project_slug` did `cwd.replace("/", "-")` — it only normalized `/`, not `_` or `.`. Claude Code stores project dirs replacing **every non-alphanumeric** character with `-`:

   ```
   cwd  = /Users/me/Algo/foo_bar
   dir  = -Users-me-Algo-foo-bar      (Claude: '_' -> '-')
   slug = -Users-me-Algo-foo_bar      (old code: '_' kept)
   ```

   So for any project whose path contains `_` or `.` (worktrees included — `/.claude/` -> `--claude-`), the computed slug never matched the project directory.

2. **Non-strict resolution falls through to a global guess.** With Strategy 3 (cwd-slug match) missing, `find_current_session(..., strict=False)` fell through to Strategy 4 — *"return the most-recently-modified session across **all** projects"* — and several context-injection / write call sites resolved non-strict. PostCompact then read whichever project was most recently active.

This was deterministic for every underscore/dot-named project and triggered whenever a second session was more recently active. The same broken slug also made the behavioral-digest memory directory silently resolve to `None` (digest no-op), and the PreCompact `checkpoint` write path had the same non-strict resolution (write-side of the same chain).

## Fix

- **`session.py` `cwd_to_project_slug`** → `re.sub(r"[^a-zA-Z0-9]", "-", cwd)`, matching Claude Code's normalization exactly (verified against real `~/.claude/projects/` directory names: underscore, dot/worktree, space, hyphen, plain — all reproduced 1:1, case and digits preserved).
- **`session.py` Strategy 3** → exact-match (`s["project"] == slug`) instead of substring, eliminating prefix collisions (`-x-foo` ⊂ `-x-foobar`).
- **Strict resolution on every cross-project-sensitive call site** (`find_current_session(cwd, strict=True)`): `cmd_post_compact` (recovery read), `_digest_session` (digest flush/inject write), `checkpoint_team` (PreCompact write), `start_guard_daemon` and `reload_self_daemon` (daemon lifecycle). On an ambiguous resolution these now return `None` / fall back to the local project instead of guessing across projects. The daemon paths are behavior-preserving (they degrade to the pre-existing no-session-id fallback).
- **`digest.py` `_get_memdir`** routed through the corrected `cwd_to_project_slug` with exact-match (it had its own independent, identically-broken inline slug derivation — not fixed transitively).
- **`team.py` `read_team_checkpoint`** gains an `include_global: bool = True` parameter (backward-compatible); PostCompact recovery now passes `include_global=False` so it never falls back to the shared global `~/.claude/team-checkpoint.md` on the recovery path.

No new dependencies — `re`/`os` are stdlib; the package stays zero-dependency.

## Tests

- **Slug parity** (table-driven): underscore, dot/double-dash, worktree, space, unicode, digits, existing-hyphens, trailing-slash — each asserted against the literal directory name Claude Code creates.
- **Cross-project isolation**: `cmd_post_compact` and `checkpoint_team` with two projects (the other more recently active) never resolve/return/​write the wrong project; positive resolution to the correct project is covered too.
- **Strategy-4 regression**: `find_current_session(cwd, strict=True)` returns `None` (not most-recent) when no slug matches.
- **Digest**: `_get_memdir` resolves an underscore project (was `None`) and never prefix-collides.
- **Daemon strict→None**: `start_guard_daemon` / `reload_self_daemon` degrade gracefully with no behavior change.

Bug-capture tests were confirmed RED against the base commit before the fix. Full suite: **979 passed, 5 skipped, 0 failed** (excluding the 3 pre-existing `test_model_detection` failures unrelated to this change).

An end-to-end before/after reproduction of the original scenario confirms it: against the unpatched code, `cmd_post_compact` for an underscore project emitted the *other* project's checkpoint; after the fix it emits the correct project's checkpoint (or nothing), the resolver returns the correct project, and the digest memory directory resolves.

## Scope (intentionally not changed)

- `project_slug_to_path` (`slug.replace("-", "/")`) — pre-existing, documented lossy round-trip; the authoritative sidecar CWD fallback already covers it.
- `cmd_current` — interactive display; a wrong guess is visible to the user, so non-strict is appropriate.
- `doctor.py` substring matches — those slugs come from guard PID-file stems (session-id namespace), not the cwd project slug.

## Deployment note

Merging is necessary but not sufficient for the fix to take effect for an installed user: it lands once a new release is published to PyPI (or via `pip install --upgrade cozempic`).

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -746,8 +746,11 @@ def cmd_post_compact(args):
 
     cwd = args.cwd or os.getcwd()
 
-    # Try to find project dir from current session
-    sess = find_current_session(cwd)
+    # Try to find project dir from current session.
+    # strict=True: refuse to fall back to the global most-recent session (Strategy 4).
+    # If no session is found, use Path(cwd) — reads from the local project dir,
+    # which is always project-correct and never cross-contaminates.
+    sess = find_current_session(cwd, strict=True)
     project_dir = Path(sess["path"]).parent if sess else Path(cwd)
 
     content = read_team_checkpoint(project_dir)
@@ -1105,7 +1108,9 @@ def _digest_session(args):
         # consistent with how the rest of cli.py resolves the current session.
         # Routing "current" through resolve_session() would use process-detection
         # (find_current_session(strict=False)) which is a different strategy.
-        sess = find_current_session(cwd)
+        # strict=True: digest flush/inject are write operations — injecting rules
+        # into an unrelated session's JSONL is a context-contamination risk.
+        sess = find_current_session(cwd, strict=True)
         if not sess:
             # Changed to stderr: consistent with resolve_session error output
             # and all other error paths in cli.py.
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -746,14 +746,17 @@ def cmd_post_compact(args):
 
     cwd = args.cwd or os.getcwd()
 
-    # Try to find project dir from current session.
-    # strict=True: refuse to fall back to the global most-recent session (Strategy 4).
-    # If no session is found, use Path(cwd) — reads from the local project dir,
-    # which is always project-correct and never cross-contaminates.
+    # Resolve the project dir for this session.
+    # strict=True: refuse Strategy 4 (global most-recent fallback) — wrong-project
+    # sessions produce wrong-project checkpoints.
+    # On None: Path(cwd) points to the local project dir (project-correct; may have
+    # no checkpoint file → read_team_checkpoint returns None → silent, which is safe).
     sess = find_current_session(cwd, strict=True)
     project_dir = Path(sess["path"]).parent if sess else Path(cwd)
 
-    content = read_team_checkpoint(project_dir)
+    # include_global=False: the global ~/.claude/team-checkpoint.md is a cross-project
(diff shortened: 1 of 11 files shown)


junaidtitan:
Thanks @ynaamane — sharp catch. The slug bug (only `/` normalized, not `_`/`.`) was deterministically injecting the *wrong project's* team-checkpoint after compact for any underscore/dot path, and silently no-op'ing the digest memdir for those projects too. `re.sub(r'[^a-zA-Z0-9]', '-', …)` matches Claude Code's dir naming 1:1, the exact-match Strategy 3 kills the prefix-collision, and flipping the cross-project-sensitive call sites to strict resolution is right. Reviewed: 1029 tests, RED-at-base verified, and strict→None degrades gracefully on every flipped site (guard/treat/reload still resolve). Merging.

Issue #106, opened 2026-05-29
Guard daemon rewrites the live transcript out from under Claude Code (data-loss race)


AndrewChemis:
## Problem

The guard daemon mutates the live session JSONL — the file Claude Code holds open and is actively appending to — by writing a temp file and `os.replace`-ing it (`session.py:732`, via `guard_prune_cycle` → `save_messages`). This makes cozempic a second, uncoordinated writer to a file the harness owns.

`_FileSnapshot` (`session.py:37`) and `_PruneLock` defend against *other cozempic processes* and detect appends that landed **before** `classify()`. They cannot close two real gaps:

1. **TOCTOU window** between `snapshot.classify(path)` (`session.py:700`) and `os.replace` (`session.py:732`) — lines the harness appends in that window are lost.
2. **Inode swap** — after `os.replace`, any write buffered in the harness's open fd targets the now-unlinked old inode and is silently dropped. Locking can't fix this because the harness isn't party to the lock.

This manifests as truncated / duplicated / interleaved transcript content.

## Impact (why this matters)

Corrupting the transcript corrupts what the agent reads back as its own history and tool output. In one session, the resulting garbled tool output led the agent to conclude there was a **severe security issue that didn't exist** — it escalated a phantom threat and **paused all work**. The failure mode isn't just lost text: it feeds the model false context, and the model acts on it. A context-protection tool that can make the agent hallucinate a security incident and halt is doing net harm.

## Evidence

Normal session, 1M-token context window. The SOFT tier fires at **25%** context and re-checks every 30s, producing repeated live `os.replace`es:

- `.jsonl.bak` files + guard logs show **3 rewrites of an actively-running session within ~3 minutes**.
- An older session logged 165 cycles, most freeing 0 bytes — i.e., re-locking, re-reading, and replacing a live file repeatedly for no benefit.

## Open design question

Does rewriting the on-disk JSONL even shrink the **live** context? Those tokens were already sent to the API and the harness holds its own in-memory state; the rewrite plausibly only affects `--resume`. If so, live-session rewriting is risk with no upside.

## Recommendation

- Only prune at boundaries where Claude isn't holding the file — as part of the terminate → `--resume` path, and via the `PreCompact` hook. Keep continuous checkpointing read-only (`load_messages_incremental`); never `os.replace` a live, actively-appended file.
[...]

PR #114, opened 2026-06-02
fix(safety): prune-safety defense-in-depth — post-prune validation + floor + metadata protection (#102 follow-up)

ynaamane:
## Problem

Follow-up to #102 (closed). v1.8.18's terminate-first fix resolved the #106 write-timing race; this PR adds the **orthogonal defense-in-depth layer** you invited — post-prune structural validation, floor preservation, and last-of-type metadata protection — none of which touch the terminate-first / idle-guard logic 1.8.18 owns.

Three gaps remained after 1.8.18:
1. **`permission-mode` dropped on compact-collapse** — `gentle/compact-summary-collapse`'s `_META_TYPES` set omitted `permission-mode`, so collapsing a session silently dropped the entry Claude Code needs to bootstrap resume (the original 63MB→730KB unresumable-session root cause).
2. **No post-prune structural validation** — a strategy could produce a structurally unreplayable session and `save_messages` would persist it.
3. **No floor** — a removal strategy could drop ~100% of conversation turns.

## Fix

- **P0-D — metadata singleton protection**: add `permission-mode` to `_META_TYPES`; generic last-of-type protection (`ai-title`/`last-prompt`/`permission-mode`) via a `__cozempic_metadata_singleton__` tag honored by `is_protected()`.
- **P0-B — post-prune validation** (`safety.py:validate_post_prune`, C1–C7): runs inside `run_prescription` **before** the deferred writer, so on failure the prune aborts with the file untouched and Claude **not** terminated (fully compatible with the 1.8.18 terminate-first flow). C1 (parent resolution) and C2 (root preservation) are **baseline-relative**: cross-session anchors and legitimately orphan-fixed cross-session `tool_result` shells on resumed sessions are excused; only prune-induced chain breaks raise.
- **P0-C — floor preservation** (`enforce_floor` + new `config.py`): max-drop cap (default 50%), last-K turn preservation (default **10**), first-message preservation, with paired `tool_use`/`tool_result` re-add. Floor is always-on in production (not disable-able via external config).
- Exit code 5 on `cozempic treat`/`reload` when a prune would be structurally invalid; the daemon logs + aborts the cycle (no write, no terminate).

## Tests

42 new tests (`test_prune_safety.py`, `test_permission_mode_protected.py`, `test_prune_safety_r2.py`), all proven assertion-RED at base. Full suite **1046 passed / 5 skipped / 0 failed**. The 1.8.18 live-fd contract (`test_guard_live_readonly.py`) is untouched and green. Zero new dependencies (stdlib only).

## Scope

- **Not included** (1.8.18 owns them): the P0-A active-session idle guard and the terminate-first / `_terminate_and_resume` restructure from #102.
- **Behavior change**: `preserve_last_k_turns` defaults to **10** (vs #102's 50 — 50 neutralized removal strategies on ≤50-turn sessions; 10 keeps a meaningful guarantee without no-op'ing small-session pruning). Override via `COZEMPIC_FLOOR_PRESERVE_LAST_K`. Other env vars: `COZEMPIC_FLOOR_MAX_DROP_PCT`, `COZEMPIC_FLOOR_PRESERVE_FIRST`.
- **Deferred**: the K=10 daemon diagnostic could distinguish validation-failure-dominated cycles from live-context dominance (follow-up). Two LOW theoretical residuals noted in review (the C1 orphan-shell skip narrows a fallback that relink already covers; `_build_orphan_shells` 2-pass cost, negligible under the 200MB size-guard).

Built via a planner→builder→reviewer pipeline with two adversarial review rounds + a final max-effort multi-angle pass:
- **Round 1** caught a CRITICAL — orphan-fix-after-floor re-dropping a re-added cross-session root → C2 abort on resumed sessions — plus a tautological abort-contract test (the mock target didn't bind through `from .executor import`); both fixed and mutation-verified.
- **Round 3 polish**: `config.py` now parses `preserve_first_message` from a JSON config file via `_parse_bool` (a string `"false"`/`"0"` was being read as `True`, inconsistent with the env-var path), and the team-protect tag-apply was moved inside the `try/finally` so the strip is unconditional.

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -9,15 +9,16 @@ import subprocess
 import sys
 from pathlib import Path
 
+from .config import load_config
 from .diagnosis import diagnose_session
 from .doctor import run_doctor
 from .executor import execute_actions, run_prescription
 from .guard import checkpoint_team, start_guard, start_guard_daemon
-from .safety import PruneValidationError
+from .helpers import is_ssh_session, shell_quote
 from .init import run_init
 from .recap import save_recap
 from .registry import PRESCRIPTIONS, STRATEGIES
-from .helpers import is_ssh_session, shell_quote
+from .safety import PruneValidationError
 from .session import _PruneLock, PruneConflictError, PruneLockError, find_claude_pid, find_current_session, find_sessions, get_session_cwd, load_messages, project_slug_to_path, resolve_session, save_messages, snapshot_session
 from .tokens import estimate_session_tokens, quick_token_estimate, calibrate_ratio
 from .types import PrescriptionResult, StrategyResult
@@ -252,9 +253,12 @@ def cmd_current(args):
         print_diagnosis(diag, sess["path"])
 
         print("  Estimated Savings by Prescription:")
+        # Resolve floor_config once (L-3: avoid per-prescription disk reads in the loop)
+        _floor_cfg = load_config().floor
         for rx_name, strategy_names in PRESCRIPTIONS.items():
             try:
-                new_msgs, _ = run_prescription(messages, strategy_names, {})
+                new_msgs, _ = run_prescription(messages, strategy_names, {},
+                                               floor_config=_floor_cfg)
             except PruneValidationError as ve:
                 # Dry-run estimation: a structural validation failure means the
                 # session is already malformed or the strategy would make it
@@ -276,9 +280,12 @@ def cmd_diagnose(args):
     print_diagnosis(diag, path)
 
     print("  Estimated Savings by Prescription:")
+    # Resolve floor_config once (L-3: avoid per-prescription disk reads in the loop)
+    _floor_cfg = load_config().floor
     for rx_name, strategy_names in PRESCRIPTIONS.items():
         try:
-            new_msgs, _ = run_prescription(messages, strategy_names, {})
+            new_msgs, _ = run_prescription(messages, strategy_names, {},
+                                           floor_config=_floor_cfg)
         except PruneValidationError as ve:
             # Dry-run estimation: report structural failure instead of crashing.
             check = ve.evidence.get("failed_check", "?")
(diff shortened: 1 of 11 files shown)


junaidtitan:
Merged into `main` — thank you, @ynaamane. 🙏

This is exactly the defense-in-depth layer we invited after #102, and it landed with real rigor. We put it through a deep multi-lens review (correctness, false-positive/guard-integration, and protection-completeness), and the core held up beautifully:

- **C1 is correctly baseline-relative** — it only flags prune-induced chain breaks and properly skips cross-session parentUuid pointers. That was the subtle part of the #102 feedback and you nailed it.
- **C2 is orphan-shell-aware** and closes the relink-to-null fake-root bypass.
- **C5 catches the original root cause** — the dropped `permission-mode` that made sessions unresumable — which is the whole reason this work mattered.
- **P0-D singleton protection is honored across all strategies**, the `__cozempic_metadata_singleton__` tag never leaks to disk (verified even on the validation-raise path), and **guard integration is safe** — a rejected prune is treated as a no-op that feeds the circuit breaker and backs off, never a crash or a kill-then-fail-resume.
- Floor logic is correct and the whole validation runs O(n) (~11–20ms on 5K-message sessions).

One gap the review surfaced, which we added as a small follow-up commit on top of your branch (`817cbb8`): a **C8 check** — a surviving `tool_use` whose paired `tool_result` was pruned passes C1–C7 but is unresumable (the API rejects a `tool_use` with no matching `tool_result`), the mirror of the orphaned-`tool_result` case. C8 is baseline-relative just like C1, so an in-flight final-turn `tool_use` that never had a result still passes. Tests cover reject / intact-pair / in-flight / both-halves-dropped.

Genuinely strong, safety-critical work — this materially hardens the prune path against the exact class of failure that started all this. Thank you for resubmitting it and seeing it through. Looking forward to your next one. 🚀

(Merged via `6b3ff2c` → `main` at `817cbb8`; closing this PR as merged since the fork→merge-commit path didn't auto-flag it.)

PR #116, opened 2026-06-07
fix(validation): reject NaN/inf and non-representable ints in positive-number coercion

ynaamane:
## Problem

`coerce_positive_float` (and its siblings) gate positivity with `value <= 0`. By IEEE-754, **every comparison with `NaN` is `False`**, and `inf <= 0` is also `False` — so `NaN`/`inf` slip past the guard and are returned/used as if valid. This silently poisons downstream numeric thresholds (a `NaN` threshold compares `False` to everything, effectively disabling the gate it controls).

The bypass exists at every place user-supplied positive numbers are coerced:

- **Config-dict path** — `_validation.coerce_positive_float`. Reachable via `~/.cozempic/config.json`: CPython's `json.loads` accepts the bare `NaN`/`Infinity` tokens by default, so `{"...": NaN}` flows a real `NaN` into a strategy/threshold config.
- **CLI path** — `cli._positive_float` (argparse `type=` for `--threshold` / `--soft-threshold`). `float("inf")`/`float("nan")` parse fine, so `cozempic guard --threshold inf` silently sets a non-finite MB threshold.
- **Guard daemon validators** — `start_guard`, `start_guard_daemon`, and `reload_self_daemon`. These belt-and-braces checks exist specifically to validate **direct Python callers that bypass argparse** (per their own docstrings), but `nan <= 0` / `nan >= hard` defeat them. `reload_self_daemon` is the worst case: it SIGTERMs/SIGKILLs the live daemon **first**, then would raise downstream — orphaning the session with no protection.

A non-representable huge `int` (e.g. `10**400`) is the same class: it survives the numeric type guard, then raises a bare `OverflowError` deep in arithmetic instead of a clean `ConfigError`.

## Fix

Reject non-finite (and non-representable) values **before** the positivity check, with a clear `"must be a finite number"` error — mirroring the existing `config.py:_clamp_float` convention. Zero new dependencies (`math.isfinite`, stdlib).

- `_validation.coerce_positive_float` → `ConfigError` for `NaN`/`inf`/huge-int.
- `cli._positive_float` → `argparse.ArgumentTypeError` for `NaN`/`inf`.
- `guard.py` → one shared `_validate_finite_thresholds(...)` helper (DRYs two previously byte-identical loops), called at the **top** of `start_guard`, `start_guard_daemon`, and `reload_self_daemon` — i.e. **before** `reload_self_daemon`'s destructive kill, so an invalid config can never orphan a live daemon.

## Tests

- Strict TDD: every bug-capture test proven RED at the PR base (returns `nan`/`inf` / raises `OverflowError`) before the fix, GREEN after.
- Daemon tests are isolated **by design** (mocked spawn claim + `mkdtemp`/teardown) — verified to create **zero** real `/tmp/cozempic_guard_*` pidfiles even when run against unpatched source.
- Full suite: **1104 passed, 0 failed, 5 skipped**.

## Scope

In scope: the positive-number coercion `NaN`/`inf`/huge-int bug class, at all four entry points.

Explicitly out of scope (intentional or pre-existing, different bug class):
- `config.py:_clamp_float` uses a clamp-to-default shape (degrade gracefully) vs the raise shape here — deliberate, not merged.
- Accepting `threshold_mb=True` (bool) — pre-existing type-validation gap, separate concern.
- numpy / `Decimal` numeric types — cozempic is zero-dependency; argparse/JSON produce builtin `float`/`int`, and there are no such callers.

### DIFF
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -24,6 +24,7 @@ Design goals:
 
 from __future__ import annotations
 
+import math
 import os
 import sys
 from typing import Any
@@ -111,7 +112,7 @@ def coerce_positive_float(config: dict, key: str, default: float) -> float:
     """Return `config[key]` as a strictly-positive float, or `default` if absent.
 
     Accepts int in addition to float (users write `threshold=50`). Rejects
-    bool, str, None, negative, and zero.
+    bool, str, None, negative, zero, NaN, and infinity.
     """
     if key not in config:
         return default
@@ -119,6 +120,10 @@ def coerce_positive_float(config: dict, key: str, default: float) -> float:
     if not _is_strict_number(value):
         raise ConfigError(
             f"config[{key!r}] must be a number, got {type(value).__name__} {value!r}"
+        )
+    if math.isnan(value) or math.isinf(value):
+        raise ConfigError(
+            f"config[{key!r}] must be a finite number, got {value!r}"
         )
     if value <= 0:
         raise ConfigError(
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -122,7 +122,7 @@ def coerce_positive_float(config: dict, key: str, default: float) -> float:
             f"config[{key!r}] must be a number, got {type(value).__name__} {value!r}"
         )
     try:
-        _nonfinite = math.isnan(value) or math.isinf(value)
+        _nonfinite = not math.isfinite(value)
     except OverflowError:
         # int too large to convert to float (e.g. 10**400) — treat as non-finite
         _nonfinite = True
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -121,7 +121,12 @@ def coerce_positive_float(config: dict, key: str, default: float) -> float:
         raise ConfigError(
             f"config[{key!r}] must be a number, got {type(value).__name__} {value!r}"
         )
-    if math.isnan(value) or math.isinf(value):
+    try:
+        _nonfinite = math.isnan(value) or math.isinf(value)
+    except OverflowError:
+        # int too large to convert to float (e.g. 10**400) — treat as non-finite
+        _nonfinite = True
+    if _nonfinite:
         raise ConfigError(
             f"config[{key!r}] must be a finite number, got {value!r}"
         )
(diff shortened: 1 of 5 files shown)


junaidtitan (approved):
Merged and **shipped in [v1.8.23](https://github.com/Ruya-AI/cozempic/releases/tag/v1.8.23)** 🙏 — thank you, @ynaamane.

This is exactly the kind of fix that's kept cozempic's guard as solid as it is: a real, silent, IEEE-754-shaped bug (`NaN`/`inf` sliding past `value <= 0`, a `NaN` threshold then comparing `False` to everything and quietly *disabling* the gate it controls), diagnosed precisely across all four entry points — config, CLI, and the daemon validators **before** the destructive `reload_self_daemon` kill so an invalid config can't orphan a session — with strict TDD (red→green) and clean, dependency-free `math.isfinite` guards. It applied cleanly on top of the 1.8.22 interactive-guard work with zero conflicts.

Two things we did on the back of it, both because of your work:

1. **We folded in the sibling your fix pointed us at.** Running the same lens over the new 1.8.22 code, the *exact* class had recurred in a knob we'd just added: `COZEMPIC_RELOAD_WARN_GRACE=NaN/inf` made `elapsed >= grace` permanently `False`, silently wedging the interactive idle-reload — the very fallback that exists so a missing nudge can't wedge reloads. Now clamped, alongside `COZEMPIC_NUDGE_PCTS` and the idle-cycle caps. We'd have shipped that without #116 prompting the sweep.

2. **We made this a standing part of our own QA instead of relying on you to catch it.** New `tests/test_input_coercion_corpus.py` throws an adversarial corpus (`nan`/`inf`/`1e999`/huge-int/unicode/empty/type-confusion) at *every* numeric validator, and `docs/qa-fleet.md` now encodes the full adversarial fleet — every bug class you and other contributors have surfaced (input validation, daemon concurrency, resource exhaustion, filesystem safety, cross-platform, injection, prune safety, reload safety, hook protocol, …) — to run on every full QA. The goal is to surface these ourselves before review.

You're cozempic's most prolific contributor by a wide margin, and the [CONTRIBUTORS](https://github.com/Ruya-AI/cozempic/blob/main/CONTRIBUTORS.md) now leads with that. Genuinely grateful for the depth and rigor you keep bringing. 🚀

PR #117, opened 2026-06-08
fix(guard): safe-point reload gate is blind to Agent-spawned agent teams

ynaamane:
## Problem

The 1.8.22 safe-point reload gate (`safe_to_reload`) is meant to never SIGKILL Claude through an active agent team — explicitly the guard's primary use case. But for a team spawned with the **`Agent` tool** (the current harness mechanism), the gate's teammate/subagent protection is **inert**, so a live team can be reloaded out from under itself:

- `extract_team_state` read `inp.get("name")` for `TeamCreate`, but the tool emits **`team_name`** → `state.team_name=""` → `merge_config_into_state` never matches the team's `config.json` → `teammates=[]`.
- `"Agent"` is not in `TEAM_TOOL_NAMES`, so `Agent`-tool teammate spawns create no teammate/subagent entry from the JSONL either.
- `detect_in_flight` does not match the `Agent`-spawn result text (`"Spawned successfully. agent_id: …"`), and that spawn already has a `tool_result`, so it is not an open call.

Net: a team coordinated via `SendMessage` with **no active shared-list task** + an idle lead → `safe_to_reload` returns `(True, "quiescent")` → the guard reloads (SIGKILL) the lead and orphans the live team. The only thing that protected such teams was an active task in the lead's transcript. Reproduced against real session transcripts.

## Fix

- **Read the real key**: `team_name` (fallback `name`), restoring config-merge so the team is visible (also fixes checkpoint accuracy).
- **Recognize `Agent`-tool teammate spawns** via an *extract-only* `_TEAM_EXTRACT_TOOL_NAMES = TEAM_TOOL_NAMES | {"Agent"}` — deliberately **not** added to `TEAM_TOOL_NAMES`, so `prune_with_team_protect` is unchanged and Agent calls are not over-preserved (no pruning regression). Placeholder `status="running"`, re-keyed to the real `agentId` parsed from the result.
- **Resolve bare-name `SendMessage(to:)`** (`"alice"`) against a `name→agentId` index (members are keyed `"alice@team"`).
- **Terminal transition**: a teammate goes benign `"idle"` on an `idle_notification` `<teammate-message>` (chronology-aware vs re-activating `SendMessage`s), and the idle-notification carrier is now prune-protected so the clear signal survives a prune (no permanent wedge).
- **Failed spawn**: an `Agent` result with no parseable `agent_id` marks the placeholder `"failed"` so a failed spawn cannot block reloads forever.
- **Teammate block fires unconditionally** on a non-benign status. A non-benign status can only originate from *this* session's JSONL; config-only members are `"config"` (benign) and never block. (An earlier session-id gate was prototyped and removed — it misfired on a reused-team-name stale `config.json`, injecting an old `leadSessionId` and skipping the block for a live team = the same catastrophe inverted.)

## Tests

- New `tests/test_guard_team_agent_spawn.py` (31 tests). Each regression guard proven RED at the PR base; characterization tests labeled as such.
- Covers: team_name extraction, Agent-spawn recognition, bare-name resolution, idle transition + chronology guard, idle-carrier prune-protection, failed-spawn cleanup, unconditional-block, stale same-name config (intersecting members) must still block, nested-`<teammate-message>` isolation.
- `docs/qa-fleet.md` L8 extended with the Agent-spawn team-visibility standing test.
- Full suite: **1225 passed, 5 skipped** (one pre-existing flaky `TestG4_PidfileWriteIsAtomic` pidfile race — untouched module, passes in isolation).
- Zero new dependencies (stdlib only — preserves the zero-dep design).

## Scope

- **Deferred** (separate follow-up, different subsystem): two LOW latent armed-sentinel items — `_write_armed_atomic` tmp-suffix uses `os.getpid()` only (narrow concurrent-teardown race), and `[:12]` session-slug collision (negligible for UUIDv4; shared module-wide convention).
- **Documented residual**: a teammate could stay `"running"` if its terminal signal is lost to asymmetric pruning → the reload defers (read-only checkpoint, autocompact fallback) rather than SIGKILLing — strictly better than the bug it replaces.

### DIFF
--- a/docs/qa-fleet.md
+++ b/docs/qa-fleet.md
@@ -76,14 +76,19 @@ validation + a floor (don't over-prune below a safe ratio). Refuse to prune an a
 in-flight session. Metadata-singleton protection. Team state survives compaction
 (checkpoint + PostCompact re-inject). *Motivating PRs: #114, #102, #22, #110.*
 
-### L8 — Reload safety (the 1.8.x guard line)  ·  standing tests: `tests/test_guard_safe_point.py`, `tests/test_interactive_*`
+### L8 — Reload safety (the 1.8.x guard line)  ·  standing tests: `tests/test_guard_safe_point.py`, `tests/test_interactive_*`, `tests/test_guard_team_agent_spawn.py`
 A reload SIGKILLs + resumes, so: NEVER reload through in-flight work (running Workflow
 / background `Agent` subagent / agent team / open tool call) — defer instead.
 Interactive: warn-before-reload, reload only at a sustained idle breakpoint, force only
 near the wall. Armed-sentinel lifecycle: cleared on **every** reload path
 (`_terminate_and_resume` choke point), atomic writes, no stale `warned` survives a
 reload. Markers must be matched to the REAL harness output (verify against live
-transcripts, not fixtures). *Motivating: the 1.8.19–1.8.23 guard line.*
+transcripts, not fixtures).
+Agent-spawn team visibility: `extract_team_state` must see teammates created via the
+`Agent` tool (not just `TeamCreate`). `safe_to_reload` must block when any teammate is
+non-benign/non-terminal, and must NOT be wedged by a stale cross-session TeamState
+(session anti-wedge). Standing test: `tests/test_guard_team_agent_spawn.py`.
+*Motivating: the 1.8.19–1.8.23 guard line; F1 fix (v1.8.24).*
 
 ### L9 — Error handling & hook protocol
 Hook commands NEVER block and ALWAYS exit 0 (a crash must not break the session); no
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -2425,6 +2425,25 @@ def detect_in_flight(messages) -> dict:
     }
 
 
+def _team_is_current_session(team_state, session_path) -> bool:
+    """True if the TeamState belongs to the session currently being guarded.
+
+    A stale config.json from a prior session can present a non-quiescent
+    team_state (teammates still "running") long after those agents exited.
+    Comparing the state's lead_session_id to the guarded session's stem prevents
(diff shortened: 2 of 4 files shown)


junaidtitan:
tested and verified

junaidtitan:
Thank you, @ynaamane — this is the kind of fix that matters most: a silent, safety-critical false-negative where an `Agent`-spawned team was invisible to the reload gate and could be SIGKILLed mid-flight. **Shipped in v1.8.24.** 🎉

While folding it in we ran the bundle through our adversarial QA fleet and hardened a few adjacent edges on top of your fix:
- **camelCase `agentId`** parsing — the shipped 1.8.22 background marker is camelCase, so a snake-only matcher would silently miss a real team.
- **`SpawnTeammate`** + a **deny-by-default net**: if the transcript shows team coordination we can't resolve to a roster, the gate now blocks rather than reloads.
- **`TeamDelete`** anti-wedge and a **fail-safe** for spawn results we can't parse (keep blocking instead of assuming failure).

In the same pass, the fleet caught — and we fixed — two over-block regressions **in our own hardening** (not in your change): the net was matching marker-shaped strings in arbitrary tool output (e.g. `agent_id:` in a grep/Read result) and protocol words in prose, which would have over-blocked teamless sessions and made the guard inert. We corrected it to correlate a spawn marker to its paired spawn tool_use and to match the *structural* `<teammate-message>` carrier on the harness delivery surface, and scoped `TeamDelete` so it can never touch another live team. Both failure directions are now pinned by a standing contract test (`tests/test_reload_gate_contract.py`).

One ask if you're up for it: a **redacted real agent-team transcript** would let us ground-truth these markers against reality (we keep matchers honest with `tests/fixtures/harness/`). No worries if not.

That's **28 merged PRs** now — crash-prevention, daemon races, PID identity, prune safety, and now agent-team reload safety. Thank you for keeping making this tool safer. 🙏

PR #122, opened 2026-06-10
feat: add --protect-pattern <regex> flag to treat, guard, reload

eggrollofchaos:
## Summary

Adds `--protect-pattern <regex>` flag to `treat`, `guard`, and `reload` commands. Messages whose text content matches any pattern get the same pruning immunity as hardcoded team-message protection (TeamCreate/SendMessage/TaskCreate).

## Motivation

Coordination-protocol blocks, control-file instructions, gate contracts, and standing obligations need to survive compaction. Currently, only hardcoded message types (team tool uses, system boundaries) are protected — user-defined high-importance messages are pruned like any other content.

## Design

- **helpers.py**: `is_protected()` now recognizes `__cozempic_pattern_protected__` tag. New functions: `tag_pattern_matches()` (tag messages before prune), `compile_protect_patterns()` (compile + validate regex).
- **cli.py**: `--protect-pattern` (action=append) on treat, guard, reload. Patterns compiled at parse time; matching messages tagged before `run_prescription`, tags stripped after.
- **guard.py**: `protect_patterns` threaded through `start_guard()`, `start_guard_daemon()`, `guard_prune_cycle()`, `prune_with_team_protect()`. Daemon serializes patterns as `--protect-pattern` CLI args to child process.
- **executor.py**: No changes needed — `is_protected()` is the single choke point used by all strategies.

## Usage

```bash
# Protect coordination-protocol messages
cozempic guard --daemon --protect-pattern 'control-file' --protect-pattern 'SPR-'

# One-shot treat with protection
cozempic treat session-id --protect-pattern 'gate-contract' --execute
```

## Behavior

- Flag is optional — omitted = no behavior change
- Invalid regex: fails early at CLI parse time with clear error
- Only `type: text` content blocks are checked
- Default off; no config file support yet (future)

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -15,18 +15,35 @@ from .diagnosis import diagnose_session
 from .doctor import run_doctor
 from .executor import execute_actions, run_prescription
 from .guard import checkpoint_team, start_guard, start_guard_daemon
-from .helpers import is_ssh_session, shell_quote
+from .helpers import (
+    is_ssh_session,
+    shell_quote,
+    tag_pattern_matches,
+    compile_protect_patterns,
+)
 from .init import run_init
 from .recap import save_recap
 from .registry import PRESCRIPTIONS, STRATEGIES
 from .safety import PruneValidationError
-from .session import _PruneLock, PruneConflictError, PruneLockError, find_claude_pid, find_current_session, find_sessions, get_session_cwd, load_messages, project_slug_to_path, resolve_session, save_messages, snapshot_session
+from .session import (
+    _PruneLock,
+    PruneConflictError,
+    PruneLockError,
+    find_claude_pid,
+    find_current_session,
+    find_sessions,
+    get_session_cwd,
+    load_messages,
+    project_slug_to_path,
+    resolve_session,
+    save_messages,
+    snapshot_session,
+)
 from .tokens import estimate_session_tokens, quick_token_estimate, calibrate_ratio
 from .types import PrescriptionResult, StrategyResult
 
 # Ensure all strategies are registered
 import cozempic.strategies  # noqa: F401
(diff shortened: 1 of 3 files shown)


junaidtitan:
Thank you, @eggrollofchaos — **shipped in v1.8.28.** 🎉 This is a genuinely useful capability, and I especially appreciated that you wrapped the tag lifecycle in `try/finally` so the internal tag can never persist into a saved session — that care made it easy to land.

To get it out quickly I rebased your feature onto current `main` and merged it that way rather than this branch — the diff here was dominated by a full-file autoformatter (black/ruff) reflow of `cli.py` and `guard.py` (~1600 of ~1700 lines), which would have been a large, conflict-prone merge surface; the actual feature was ~55 lines. The feature itself is yours and credited in the release notes + every commit.

While landing it I added a few things from an adversarial QA pass:
- **Broadened the matcher** to also scan tool_result content and user/string content — the prescriptions mostly prune *those* surfaces (tool outputs), not assistant prose, so this makes `--protect-pattern` actually protect prunable content.
- **ReDoS budget**: stdlib `re` has no step limit, and the guard daemon re-scans every cycle, so a catastrophic pattern like `(a+)+$` could freeze it. Matching now runs under a wall-clock budget (`COZEMPIC_PROTECT_MATCH_SECONDS`, default 2s) and fails open.
- **Over-protection warning** when a pattern would immunize most prunable content (so the prune isn't a silent no-op), `=`-form serialization so a dash-leading pattern survives the daemon spawn, and `thinking`-block scanning.

Shipped as `feat(prune): --protect-pattern` (617758d) + hardening (64d9650). Closing this PR since the feature is now on `main` and released — but the credit is all yours. Please do keep them coming. 🙏

PR #128, opened 2026-06-15
fix(watchdog): verify guard identity before guard-watchdog --fix SIGTERM (L2)

ynaamane:
## Problem

The v1.8.30 `guard-watchdog --fix` (`cmd_guard_watchdog`) SIGTERMs a PID read straight from a `cozempic_guard_*.pid` file, gated only on liveness (`h.pid_alive` — a bare `os.kill(pid,0)` that even treats `PermissionError` as alive), with **no identity check**. `watchdog.py` never imports `guard._is_cozempic_guard_process` — the argv-based check every other kill site in guard.py uses (guard.py:2915/3789/3799).

Confused-deputy: a guard loop-churns its log, then dies HARD (SIGKILL/OOM/power-loss) so its pidfile is never CAS-unlinked; the OS recycles that PID to an unrelated same-user process; an operator — or a cron/CI wrapper, which the cli's own comment explicitly invites — runs `cozempic guard-watchdog --fix` → it SIGTERMs the **innocent** process. The cli docstring even claimed `--fix` "SIGTERMs ... our own process — stopping a harmful loop is safe", which was unverified.

## Fix

- Add `GuardLoopHit.guard_confirmed: bool`, computed in `scan_guard_logs` for a live pid via a lazy `from .guard import _is_cozempic_guard_process` (argv check — **ps-only, no psutil**, so it delivers full value on the zero-dependency default install; avoids the PR #98 optional-dep trap). A dead/absent pid → `False` without a check.
- `--fix` now SIGTERMs **only** when `guard_confirmed`; a live-but-unconfirmed pid prints `refusing to kill` and leaves `arrested=0`, so the `sys.exit(3)` cron-alert path still fires on an un-arrested live loop. Docstring corrected to state the identity check.

## Tests

`tests/test_guard_watchdog.py::TestFixIdentityGate` — RED-at-base proven (against base source the captured stdout literally shows `→ SIGTERM sent to looping daemon pid 7777`, i.e. base KILLS the innocent pid) → GREEN after the gate; plus the positive `test_fix_kills_confirmed_guard`. **19/19** watchdog tests green.

## Scope

L2 daemon-kill safety only. The 2 `_pid_alive` liveness copies (this PR adds none, so the class-of-bug fold doesn't trigger), the SIGTERM-startup-window pidfile leak, and the hardcoded watcher-log path → `_guard_tmp_root` are tracked for a separate guard-cleanup PR.

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -979,17 +979,16 @@ def cmd_guard_watchdog(args):
         print(f"      {h.report.reason}")
         print(f"      futile={h.report.futile_cycles} total={h.report.total_prune_cycles} "
               f"backoff_max={h.report.max_backoff_s}s exit_seen={h.report.has_exit}")
-        if getattr(args, "fix", False) and h.pid_alive and h.pid:
-            if not h.guard_confirmed:
-                print(f"      → pid {h.pid} is alive but is NOT a cozempic guard "
-                      f"(recycled?) — refusing to kill")
-            else:
-                try:
-                    os.kill(int(h.pid), signal.SIGTERM)
-                    print(f"      → SIGTERM sent to looping daemon pid {h.pid}")
-                    arrested += 1
-                except (ProcessLookupError, PermissionError, OSError) as e:
-                    print(f"      ! could not signal pid {h.pid}: {e}")
+        if getattr(args, "fix", False) and h.guard_confirmed:
+            try:
+                os.kill(int(h.pid), signal.SIGTERM)
+                print(f"      → SIGTERM sent to looping daemon pid {h.pid}")
+                arrested += 1
+            except (ProcessLookupError, PermissionError, OSError) as e:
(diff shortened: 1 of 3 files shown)


junaidtitan:
Thanks @ynaamane — the **L2** finding (verify guard identity before `guard-watchdog --fix` sends SIGTERM) was spot-on and is exactly the right hardening for the daemon-asymmetry rule (never SIGKILL a process we can't prove is our guard).

It landed on `main` via the #138 Audit-P1 remediation (shipped in **1.8.33**): `watchdog.py` now carries the `guard_confirmed` gate routed through `guard._is_cozempic_guard_process` (ps/argv-based, fail-closed). A merge-tree of this branch into current `main` produces an **empty net diff**, so there's nothing left to merge here.

Closing as superseded — the fix is live and credited to your finding. 🙏

PR #129, opened 2026-06-15
fix(digest): skip sub-agent (sidechain) turns in extract_corrections (L6)

ynaamane:
## Problem

`digest.extract_corrections` scans `user` turns for correction signals and builds a `DigestRule` — which is injected as a `hard` behavioral rule into the MAIN session. It filters `_is_system_noise` but has **no `isSidechain` guard**. So a SUB-AGENT conversation's user-turn (role:user, type:text, inside a sidechain) carrying an explicit-correction phrase is mined as if it were the human's direct correction → **untrusted sub-agent content becomes a learned behavioral rule**. Sidechain turns are sub-agent scaffolding/prompts, not the user correcting Claude.

## Fix

One `if _is_sidechain(msg): continue` at the TOP of the loop (reusing the canonical `tokens._is_sidechain`), placed BEFORE the pre-window split — so neither a sidechain user-turn is mined NOR a sidechain assistant-turn pollutes `prev_assistant_text` (the `before` evidence) for a following main rule. `extract_corrections` is the sole turn-mining chokepoint (`update_digest` / `flush_digest` / `recover_digest` all funnel through it), so the one guard closes the class.

## Tests

`tests/test_digest.py::TestSidechainInjectionGuard` — RED-at-base proven (a sidechain correction turn yields a rule at base: `1 != 0`), GREEN after. Plus a **positive control** (the IDENTICAL phrase WITHOUT `isSidechain` DOES yield a rule — proving the flag, not the phrasing, is the discriminant) and a **bleed control** (a sidechain assistant turn no longer populates the next main rule's `before`). `test_digest.py` = 216/216.

## Scope

L6 untrusted-input / digest pollution. `digest.py` only, stdlib.

### DIFF
--- a/src/cozempic/digest.py
+++ b/src/cozempic/digest.py
@@ -21,6 +21,7 @@ from typing import Literal
 
 from ._validation import parse_env_bool
 from .helpers import get_content_blocks, get_msg_type, text_of
+from .tokens import _is_sidechain
 from .types import Message
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


junaidtitan:
Thanks @ynaamane — the **L6** finding (sub-agent / sidechain turns being mined into digest corrections) was correct and is fixed on `main` (shipped in **1.8.33** via #138): `digest.py` now imports `_is_sidechain` and skips sidechain turns in `extract_corrections` (digest.py:430), including dropping their assistant text from `prev_assistant_text`.

A merge-tree of this branch into current `main` is an **empty net diff**. Closing as superseded — fix is live, credited to your finding. 🙏

PR #130, opened 2026-06-15
fix(safety): enforce_floor must not fork the conversation DAG on a compacted session (L7)

ynaamane:
## Problem

On a compacted session, compact-summary-collapse relinks the `compact_boundary` entry's `parentUuid` to `None` (it becomes the single root — the pre-boundary history has been replaced by the compact summary). Then `enforce_floor`'s `preserve_first_message` re-adds the original first message (`root-0`, also `parentUuid=None`) → the conversation DAG now has **two roots**. `validate_post_prune` misses it: C1 skips `None` parents; C2 only checks "≥1 original root survives" (which holds — root-0). A forked DAG can leave Claude Code unable to resume coherently. (Reproduced: 1 root → 2 roots `['root-0','cb-1']`.)

## Fix (defense-in-depth)

- **P0-A (prevention):** `enforce_floor` skips re-adding any pre-boundary message when an active `compact_boundary` exists — across ALL re-add paths: preserve-first-message, last-K, survival-cap, **and** step-3 tool_use/tool_result pair-closure (the last via the same `_is_pre_boundary` gate). This respects compact-summary-collapse: the summary IS the preserved pre-boundary representation.
- **C2 adjustment:** C2's "an original root must survive" eligible set now excludes pre-boundary roots when an active boundary exists — otherwise C2 false-positives on the *legitimate* collapse (which drops `root-0`). On a compacted session root-protection shifts to **C4** ("last compact_boundary must survive"); a genuine non-compacted root-drop is still caught by C2 (verified by construction in review).
- **P0-B (net):** new **C9** in `validate_post_prune` — baseline-relative multi-root guard (`raise if roots_after > roots_before`), so the prune fails-closed if any path ever introduces a new root. Placed after C2, before C4.

## Tests

`tests/test_prune_safety.py::TestFloorCompactedSession` — RED-at-base (the 2-root fork is live + C9 absent), the exact end-to-end acceptance `test_run_prescription_compacted_session_single_root`, a C9 false-positive guard (a legitimately-resumed 2→2-root session passes), and the `hasPreservedSegment` edge. **57** prune-safety + golden-corpus tests green. (A dedicated regression test for the step-3 pair-closure gate is being added.)

## Scope

L7 prune-safety / data-integrity. `safety.py` only.

### DIFF
--- a/src/cozempic/safety.py
+++ b/src/cozempic/safety.py
@@ -631,12 +631,18 @@ def enforce_floor(
                 if btype == "tool_use":
                     tid = block.get("id", "")
                     for p in tool_use_id_to_results.get(tid, set()):
-                        if p not in must_preserve:
+                        # M-1 (review): don't re-add a PRE-boundary pair partner past an
+                        # active compact_boundary — else step-3 pair-closure re-introduces a
+                        # pre-boundary root (the 2-root fork P0-A prevents elsewhere; C9 would
+                        # otherwise have to abort the prune).
+                        pe = before_by_uuid.get(p)
+                        if p not in must_preserve and not (pe and _is_pre_boundary(pe[0])):
                             new_additions.add(p)
                 elif btype == "tool_result":
                     tid = block.get("tool_use_id", "")
                     for owner in tool_use_id_to_owner.get(tid, set()):
-                        if owner not in must_preserve:
+                        oe = before_by_uuid.get(owner)
+                        if owner not in must_preserve and not (oe and _is_pre_boundary(oe[0])):
                             new_additions.add(owner)
         if not new_additions:
             break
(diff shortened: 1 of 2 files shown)


junaidtitan:
Thanks @ynaamane — the **L7** finding (`enforce_floor` forking the conversation DAG on a compacted session) was a genuine correctness bug and the boundary-aware approach was the right call.

The fix is already on `main` (shipped in **1.8.33** via #138): `safety.py` now tracks `_last_active_compact_boundary_idx` / `_is_pre_boundary` so the floor never re-parents across a compact boundary. This branch's base predates #138 **and** the later `hashable_str()` (R6) crash-guard hardening, so merging it as-is would conflict in `safety.py` and revert live hardening — rather than rebase a now-redundant change, closing as superseded.

The substance is live and credited to your finding. 🙏

PR #131, opened 2026-06-15
fix(guard): agents_active counts active teammates, not just subagents (L8)

ynaamane:
## Problem

`agents_active` (guard.py) was `any(s.status in ("running","unknown") for s in state.subagents)` — it iterated ONLY `state.subagents`, never `state.teammates`. An `Agent`-tool-spawned teammate lives in `state.teammates`, so a running teammate + empty subagents → `agents_active=False`. That feeds the futile-loop **K-exit deferral** (guard.py:766-794): with `agents_active=False` the deferral doesn't fire → the daemon `sys.exit`s on K consecutive empty prunes → the live teammates lose their guard daemon → context bloats → native autocompact destroys the team state (the exact failure the guard exists to prevent). `safe_to_reload` got teammate-awareness in the F1/#117 work; this K-exit path never did.

## Fix

Factored `_compute_agents_active(state)` that ORs the subagent check with a teammate check using the **same predicate** as `safe_to_reload` — byte-identical, so the K-exit and the reload gate can never disagree about "active." Widened the deferral diagnostic log to report active teammates, and `getattr(state, "teammates", None) or []` for legacy-mock back-compat.

## Tests

`tests/test_guard_team_agent_spawn.py::TestAgentsActiveTeammateBlind` (helper, 6 cases) + a **behavioural** integration test `test_guard_polish_pr93.py::...test_k10_defers_when_teammate_running`: a running teammate with empty subagents DEFERS the K=10 exit — behaviourally RED at base, GREEN after. Benign/idle/terminal/unknown teammate → agents_active stays False so a genuinely-quiescent team still K-exits.

## Scope

L8 reload-safety, guard.py + team.py, stdlib only.

---

An independent max-recall review upgraded a PLAUSIBLE finding to a real safety bug, fixed here; an adversarial review (no CRITICAL/HIGH) confirmed the direction-safety:

- **Allowlist/denylist asymmetry closed (the headline).** Even after the teammate fix above, `_compute_agents_active` still used an **ALLOWLIST** for *subagents* (`s.status in ("running","unknown")`) while `safe_to_reload` uses a **DENYLIST** (`not in _STATUS_TERMINAL`). They diverge for off-vocabulary / null statuses — empirically `None`, `''`, `'busy'`, `'in-progress'`, `'executing'` → `_compute_agents_active` returned False (K-exit NOT deferred) while `safe_to_reload` would block. A subagent mid-execution with such a status could lose its guard. Unified behind a single `_is_active_subagent` (DENYLIST) so the two gates can never disagree; the change only ever makes MORE things count active (over-defer = safe), never fewer.
- **Recovery-receipt predicate.** `build_team_recovery_receipt` used `not in {"done","completed"}` (2 elements), missing the other 14 terminal statuses (`failed`/`cancelled`/`aborted`/…) → a *failed* teammate counted active → a misleading `"partial"` verdict. Now uses the full quiescent set.
- **Predicate consolidation (this replaces the "deferred to a future sub-PR" note).** Extracted `_is_active_subagent` + `_is_active_teammate`, used consistently across the soft-K block, the hard-cap diagnostic, and `_compute_agents_active`. The hard-cap message now names the actual active population (subagents *and/or* teammates) instead of unconditionally saying "Subagents". The teammate terminal set lives in a **self-contained** `_TEAMMATE_QUIESCENT` frozenset in team.py (guarded by a parity test against `guard._STATUS_TERMINAL`; a future PR consolidates both + #132's cap into `_constants.py`).

Each behavioral gap has a RED-at-base test. guard.py + team.py, stdlib only.

### DIFF
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -770,20 +770,32 @@ def start_guard(
                 ):
                     # Defer: stay alive, keep cycling at backoff cap.
                     if not deferred_exit_announced:
-                        running_count = sum(
-                            1 for s in state.subagents
+                        running_subagents = sum(
+                            1 for s in (state.subagents or [])
                             if s.status in ("running", "unknown")
                         )
+                        running_teammates = sum(
+                            1 for t in (state.teammates or [])
+                            if (t.status or "").strip().lower()
+                            not in (_STATUS_TERMINAL | _TEAMMATE_BENIGN)
+                        )
+                        running_count = running_subagents + running_teammates
                         worst_case_min = (
                             HARD_LOOP_HARD_EXIT_THRESHOLD
                             * HARD_LOOP_BACKOFF_CAP_SECONDS
                             // 60
                         )
+                        parts = []
+                        if running_subagents:
+                            parts.append(f"{running_subagents} subagent(s)")
+                        if running_teammates:
+                            parts.append(f"{running_teammates} teammate(s)")
+                        active_desc = " + ".join(parts) if parts else f"{running_count} agent(s)"
                         print(
                             f"  [{_now()}] K={consecutive_empty_hard_prunes} "
                             f"reached normal exit threshold "
                             f"({HARD_LOOP_EXIT_THRESHOLD}) but "
-                            f"{running_count} subagent(s) still active. "
+                            f"{active_desc} still active. "
                             f"Deferring daemon exit until agents quiesce "
                             f"or K reaches hard cap "
                             f"({HARD_LOOP_HARD_EXIT_THRESHOLD}, "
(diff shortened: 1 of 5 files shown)


ynaamane:
**`/code-review max` hardening pass pushed** (see the updated description).

Headline fix: closed an **allowlist/denylist asymmetry**. Even after the teammate fix, `_compute_agents_active` still used an ALLOWLIST for *subagents* (`status in ("running","unknown")`) while `safe_to_reload` uses a DENYLIST (`not in _STATUS_TERMINAL`). They diverge for off-vocabulary/null statuses — `None`, `""`, `"busy"`, `"in-progress"`, `"executing"` → `_compute_agents_active` returned False (K-exit NOT deferred) while `safe_to_reload` would block, so a subagent mid-execution could lose its guard. Unified behind a single `_is_active_subagent` (DENYLIST); the change only ever counts MORE active (over-defer = recoverable), never fewer.

Also: fixed the recovery-receipt terminal-status set (`not in {"done","completed"}` missed `failed`/`cancelled`/`aborted`/…) and consolidated the active predicates (`_is_active_subagent`/`_is_active_teammate`). Each gap has a RED-at-base test.

junaidtitan:
Thanks @ynaamane — merged to main (3e21d94), applied surgically onto current 1.8.33 (the branch conflicted in start_guard's core loop against #138's PID-identity watchdog, so I lifted your change in rather than force the 3-way merge). `_compute_agents_active` + `_is_active_subagent`/`_is_active_teammate` + the soft-K/hard-cap message updates + the `_TEAMMATE_QUIESCENT` receipt set all landed, with your tests.

One note: while wiring this I confirmed (via your idle tests) that #134's H-1 gate means an idle-notification only transitions a teammate when the carrier has a top-level `teamName` — genuine harness carriers always set it, so production teammates DO quiesce; I updated the test carriers to match that real contract. Credited to you in the commit. 🙏

PR #132, opened 2026-06-15
fix(guard,team): cap the reload-gate block-regex scans against quadratic DoS (L3)

ynaamane:
## Problem

Two task-notification block-regexes in the reload-gate + checkpoint hot path ran **uncapped** DOTALL lazy-star on attacker-sized message content — O(openers × len), freezing the 30s loop:
- `guard.py:2518` `_TN_BLOCK_RE.findall(text)`
- `team.py:869` `_TASK_NOTIF_BLOCK_RE.finditer(content)`

`recap.py` already solved this class for its own regexes (`text[:32768]` / `[:8000]`); this folds the same fix into the reload gate.

A third scan on the same `content` — `team.py:917` `_TEAMMATE_MSG_RE.finditer(content)`, a negative-lookahead tempered-greedy `((?:(?!<teammate-message).)*?)` — is **O(n) linear**, not quadratic. It is capped too, for **fail-safe consistency** with its two siblings, not for performance.

## Fix

`_RELOAD_GATE_SCAN_CAP = 65536` (64 KB, named constant) caps the input to all three scans. **Fail-safe by construction:** a marker missed past the cap → the launch stays in-flight / the teammate stays active → the gate over-**defers** the reload (recoverable), never under-blocks (which would SIGKILL live work). 64 KB is ~64× a real notification; recap uses 32 KB — the larger cap suits the reload gate's larger tool-output messages.

## Tests

`tests/test_reload_gate_regex_cap.py`:
- the two quadratic regexes: **RED-at-base proven** — degenerate ~200k-opener input, **4.68 s uncapped → bounded** capped.
- `_TEAMMATE_MSG_RE`: a **cap-boundary behaviour test** — an idle-notification past the 64 KB cap must NOT transition the teammate.
- happy-path guards confirm a real notification within 64 KB still clears / transitions.

## Scope

L3 resource-exhaustion / L0 reload-gate hot path. `guard.py` + `team.py`, stdlib only.

---

An independent max-recall review surfaced a cluster of test-quality + consistency gaps (the **PR #116 trap** — tests passing for the wrong reason — plus a duplicated constant and a regex-parity gap). All fixed here; an adversarial review (PASS WITH CONDITIONS) confirmed the trap is genuinely closed, and its one MED was folded in round-2:

- **Anti-tautology test rewrites.** The two happy-path tests passed *vacuously* — `test_detect_in_flight_real_notification_still_clears` never registered an Agent launch (so `agent` was trivially falsy regardless of the cap), and the `extract_team_state` test asserted on `state.subagents` while the spawn path populates `seen_teammates`. Rewritten to genuinely exercise the path: a `cap=1` truncation now makes the assertion FAIL (verified by flipping the cap). Both the else-branch (new subagent) and if-branch (transition existing) are covered.
- **Single source of truth for the cap.** `_RELOAD_GATE_SCAN_CAP` was duplicated as two independent literals → extracted to a new pure `cozempic/_constants.py` that guard.py + team.py both import (couple-by-construction; the guard↔team cycle forbids either owning it). An `assertIs` identity test locks it; the module is wheel-build-confirmed packaged.
- **Regex parity (class-of-bug fold).** guard's `_TN_ID_RE` and `_TN_STATUS_RE` were *strict* (`<task-id>…`, `<status>…`) while team's siblings are attribute-tolerant — so an attributed `<task-id xmlns=…>` / `<status priority=…>` in a genuine completion was matched by `extract_team_state` but **dropped** by `detect_in_flight` → the launch stuck in-flight. Both broadened to byte-match team's, each with a RED-at-base test. The broadening only affects id/status extraction *within already-block-matched notifications* → it cannot false-clear (the safe direction).
- The boundary-cap test now imports the constant from the module under test; the `_tc_result` test helper takes an explicit `tool_use_id` (no index coupling).

stdlib only.

### DIFF
--- a/src/cozempic/_constants.py
+++ b/src/cozempic/_constants.py
@@ -1,0 +1,16 @@
+"""Shared compile-time constants for cozempic components.
+
+Extracted to break circular-import cycles: guard.py imports team.py at module
+level (guard.py:189), so team.py cannot import guard.py in return.  Any constant
+that BOTH modules must agree on lives here.
+"""
+
+# Maximum characters of text fed into the DOTALL lazy-star block-regex scanners
+# in detect_in_flight (guard.py) and extract_team_state (team.py).  Both scan
+# sites MUST use the same value so the two parsers agree on the same slice of
+# each message (team.py contract: "agree on the same bytes").
+# 64K characters ≈ 64× the size of a real task-notification.  A notification
+# beyond this cap is MISSED → the launch stays "in-flight" → the gate
+# OVER-DEFERS the reload (recoverable, not UNDER-BLOCKS / SIGKILL).
+# Mirrors recap.py's own DoS guard (text[:32768] / text[:8000]).
+_RELOAD_GATE_SCAN_CAP: int = 65536
--- a/src/cozempic/_constants.py
+++ b/src/cozempic/_constants.py
@@ -1,8 +1,8 @@
 """Shared compile-time constants for cozempic components.
 
 Extracted to break circular-import cycles: guard.py imports team.py at module
-level (guard.py:189), so team.py cannot import guard.py in return.  Any constant
-that BOTH modules must agree on lives here.
+level, so team.py cannot import guard.py in return.  Any constant that BOTH
+modules must agree on lives here.
 """
 
 # Maximum characters of text fed into the DOTALL lazy-star block-regex scanners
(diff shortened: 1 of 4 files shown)


ynaamane:
**`/code-review max` hardening pass pushed** (see the updated description).

Key fixes: rewrote 2 *vacuous* happy-path tests into genuine cap guards (a `cap=1` truncation now makes the assertion FAIL — the reviewer confirmed by flipping the cap 1↔65536); extracted the duplicated `_RELOAD_GATE_SCAN_CAP` into a new pure `cozempic/_constants.py` that both modules import (couple-by-construction; the guard↔team import cycle forbids either owning it); broadened guard's `_TN_ID_RE` **and** `_TN_STATUS_RE` to byte-match team's attribute-tolerant siblings (class-of-bug fold — an attributed `<task-id …>`/`<status …>` was matched by `extract_team_state` but dropped by `detect_in_flight`).

Adversarial review: PASS WITH CONDITIONS — confirmed the anti-tautology trap is genuinely closed; the `_TN_STATUS_RE` regex-parity MED folded in round-2.

junaidtitan:
Thanks @ynaamane — strong fix. Reviewed via the fleet (the adversarial pass actually *overturned* an initial reviewer worry that this reverted #134 — it doesn't).

Your 64KB regex-DoS cap itself shipped via #134, but this PR carried two things #134 didn't, which I've landed on main (daa906a) crediting you:
- guard `_TN_ID_RE` / `_TN_STATUS_RE` made attribute-tolerant, to parity with team.py — the strict form silently dropped a `<task-id attr=...>` notification → launch wedged in-flight.
- `_RELOAD_GATE_SCAN_CAP` moved to a new `_constants.py` single source of truth.

Your test suite came too. Three cases that asserted transitions from *plain* user content were re-pointed onto #134's current surfaces (task-notifications via queue-operation per C-2; idle-notifications with a top-level `teamName` per H-1) in 54ae92f, so the create/transition coverage is preserved rather than dropped. Closing as landed-via-main. 🙏

PR #133, opened 2026-06-15
test(isolation): make session-detection tests hermetic against a live ~/.claude (L11)

ynaamane:
## Problem

Three test modules resolve the current session / checkpoint through the **real `~/.claude`**, so they FAIL whenever pytest runs on a machine with an active Claude session (i.e. anyone developing cozempic with Claude Code open):

- `test_find_current_session.py` — the suite already mocked Strategy 2 (`_session_id_from_process`) and Strategy 4/5 (`get_projects_dir`), but **missed Strategy 1**: `lookup_active_transcript` reads `~/.claude/cozempic-active-sessions.json` keyed by `find_claude_pid()`. A live Claude PID resolves to the running session file, so Strategy 1 wins and returns the wrong session — preempting the Strategy 3/4 resolution the tests assert on.
- `test_guard_hardening.py::TestCheckpointTeamWriteSideIsolation`
- `test_post_compact.py::TestCmdPostCompactCrossProjectIsolation`

Reproduced against a real `~/.claude` (31 project dirs): the un-mocked tests = **10 failed / 25 passed**.

## Fix

Add `patch("cozempic.session.find_claude_pid", return_value=None)` to each affected `with` block — Strategy 1's active-transcript lookup returns `None` and falls through to the already-isolated strategies (mocked projects dir). Surgical: blocks **only** the contamination vector; the Strategy 3/4 logic under test still runs over a `TemporaryDirectory`. Test-only — no `src/` change.

## Tests

After the fix: `test_find_current_session` + `test_post_compact` = **35 passed**; `test_guard_hardening` = **94 passed**; `test_post_compact` alone = **15 passed** (robust, not order-dependent). **RED-at-base proven**: 10 failed at `origin/main` vs the same populated `~/.claude`.

## Scope

L11 test-isolation. Test files only — `test_find_current_session.py`, `test_guard_hardening.py`, `test_post_compact.py`.

---

## Follow-up — `/code-review max` hardening (still test-only)

An independent max-recall review of this branch surfaced more test-hygiene gaps in the same files; all fixed here, then an adversarial review (PASS WITH CONDITIONS) whose 2 MED conditions were folded in:

- **Two more contaminated tests patched.** `test_falls_back_safely_when_no_session_found` + `test_global_checkpoint_not_read_when_local_absent` relied on the empty-projects-dir *accident* (no explicit `find_claude_pid→None`) — fragile against future strategy reordering. Patched + a `TestPostCompactStrategy1Isolation` guard (`assertNotIn("PROJ_B_STATE")`, **RED-at-base**, GREEN after the patch) that proves the leak. *(This actually fixed ~10 of the suite's isolation failures.)*
- **Stale strategy numbering swept.** After an active-transcript strategy was inserted upstream, ~10 inline comments + the `TestStrategy3ExactMatch` class (which tests Strategy **4** = CWD-slug) were stale → renamed/corrected across the 3 files.
- **Leaked `mkdtemp` dirs cleaned.** `TestCmdPostCompactCrossProjectIsolation` left 3 temp dirs; added `addCleanup(shutil.rmtree, …)` + dropped a dead `tmp_path=None` param.
- **DRY'd 3 inline slug re-implementations.** `test_post_compact.py` + `test_guard_hardening.py` each re-implemented the cwd→slug formula with `re.sub(r"[^a-zA-Z0-9]", "-", cwd)` (no `os.path.normpath` → a trailing-slash cwd produced a wrong `-…-proj-` slug, diverging from production); replaced with the canonical `cwd_to_project_slug`. A characterization test documents the normpath contract.
- Review round-2/3 removed an always-green "gap-proof" anti-pattern test (it asserted the *bad* behavior) and re-labeled the slug test honestly (no false "RED at base" on a pure call-site DRY swap).

### Note on the one remaining full-suite failure

`tests/test_active_session_detection.py::test_two_terminals_two_pids_resolve_independently` fails on this branch — but it is **already fixed on `main`** by #124 (`05d019b`, which wraps the `record_active_transcript` calls in the `_pid_alive→True` patch). This branch's base predates #124 and does **not** touch that file, so the failure **resolves on merge** (main's fixed version is preserved). No change made here to avoid a duplicate/conflicting fold.

Test-only; `git diff origin/main...HEAD -- src/` is empty.

### DIFF
--- a/tests/test_find_current_session.py
+++ b/tests/test_find_current_session.py
@@ -28,6 +28,9 @@ class TestStrictMode:
         with (
             patch("cozempic.session.get_projects_dir", return_value=tmp_path / "projects"),
             patch("cozempic.session._session_id_from_process", return_value=None),
+            # Block Strategy 1 (active-transcript lookup keyed by live Claude PID)
+            # so a real running Claude session in the developer's home cannot win.
+            patch("cozempic.session.find_claude_pid", return_value=None),
         ):
             result = find_current_session(cwd="/unrelated/path", strict=True)
 
@@ -41,6 +44,7 @@ class TestStrictMode:
         with (
             patch("cozempic.session.get_projects_dir", return_value=tmp_path / "projects"),
             patch("cozempic.session._session_id_from_process", return_value=None),
+            patch("cozempic.session.find_claude_pid", return_value=None),
         ):
             result = find_current_session(cwd="/unrelated/path", strict=False)
 
@@ -48,7 +52,7 @@ class TestStrictMode:
         assert result["session_id"] == "aaaa1111-0000-0000-0000-000000000000"
 
     def test_strict_succeeds_when_process_detected(self, tmp_path):
-        """Process-based detection (Strategy 1) satisfies strict mode."""
+        """Process-based detection (Strategy 2) satisfies strict mode."""
         session_id = "bbbb2222-0000-0000-0000-000000000000"
         proj = tmp_path / "projects" / "-some-path"
         _write_session(proj, session_id)
@@ -56,6 +60,8 @@ class TestStrictMode:
         with (
             patch("cozempic.session.get_projects_dir", return_value=tmp_path / "projects"),
             patch("cozempic.session._session_id_from_process", return_value=session_id),
+            # Block Strategy 1 so this test exercises Strategy 2 exclusively.
+            patch("cozempic.session.find_claude_pid", return_value=None),
         ):
             result = find_current_session(strict=True)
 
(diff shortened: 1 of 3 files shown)


ynaamane:
**`/code-review max` hardening pass pushed** (test-only — see the updated description).

Key fixes: explicit `find_claude_pid→None` on 2 more tests that previously relied on the empty-projects-dir accident; swept stale strategy numbering (incl. a `TestStrategy3→4ExactMatch` rename); DRY'd 3 inline `_correct_slug` re-implementations to the canonical `cwd_to_project_slug` (the inline copies lacked `os.path.normpath`); removed an always-green anti-pattern test.

Note: the one remaining full-suite failure, `test_two_terminals_two_pids_resolve_independently`, is **already fixed on `main` by #124** (`05d019b`) — this branch predates it and does not touch that file, so it resolves on merge (no duplicate fold made here). Adversarial review: PASS WITH CONDITIONS, both MEDs addressed.

PR #135, opened 2026-06-15
fix(reload_lock): lowercase _slug_for for cross-producer slug parity (L5 split-brain)

ynaamane:
## Problem — slug split-brain

`reload_lock._slug_for` was the **only** slug producer that did not lowercase its input:

| producer | behaviour |
|---|---|
| `reload_lock._slug_for` | `_SAFE_CHARS_RE = [^a-zA-Z0-9_-]`, **no `.lower()`** |
| `spawn_lock._slug_for` | lowercases first |
| `guard._reload_armed_path` | lowercases first |
| SessionStart hook sanitiser | `re.sub(r'[^a-z0-9_-]', '_', s.lower())` |

For any uppercase-containing `session_id`, the producers emit **different** slugs → the reload lock lives at a different path than the pid / sentinel / armed files → the concurrency guard is silently bypassed. This violates the co-location invariant documented at `reload_lock.py:78` and `spawn_lock.py:205` (*"Same session_id → same slug across all three"*) — `spawn_lock`'s docstring asserted the very parity that was broken.

Practical trigger is nil today (session IDs are lowercase UUIDs), so this is **latent / defensive** — but it's a real cross-module desync plus a false docstring, and the slug functions also accept path-derived inputs which on case-preserving filesystems could carry uppercase.

## Fix

`reload_lock._slug_for` now lowercases before substitution (`session_id.lower()`), with its char class tightened to `[^a-z0-9_-]` (accurately reflecting what it matches post-lowercase — functionally identical, since uppercase is already gone). Docstrings corrected on both `reload_lock._slug_for` and `spawn_lock._slug_for`.

## Tests — `tests/test_reload_lock_wave2.py`

- `test_slug_parity_three_producers_agree_on_uppercase_input` — **RED-at-base** (`'ABCD1234EFGH' != 'abcd1234efgh'`) → GREEN (reload == spawn == guard slug).
- `test_slug_parity_uuid_input_unchanged` — passes at base **and** after the fix (zero regression for the lowercase-UUID inputs that actually occur).
- 17/17 reload_lock tests green.

## Scope

L5 cross-platform / slug co-location parity. `reload_lock.py` + `spawn_lock.py` (docstring) + tests. stdlib only.

---

An independent max-recall review of this branch surfaced a cluster of related latent gaps; all fixed here (one commit per pillar, each RED-at-HEAD where behavioral), then an adversarial review (min-10 findings, 3 gates: pre-existing-test / fold-invariant / zero-dep) returned **PASS WITH CONDITIONS** → its 2 MED conditions were folded in round-2:

- **P0-A** — `guard._reload_armed_path` + `_reload_ledger_path` now call the canonical `reload_lock._slug_for` instead of **inlining** the slug formula. The inline copies stemmed differently for path-form inputs and used a `"session"` (not `"default"`) empty-fallback — so the documented "same slug across all producers" invariant only held for UUID inputs. Calling `_slug_for` makes it hold universally (class-of-bug fold: both functions). **UUID slug-invariance preserved** — `Path(uuid).stem == uuid`, so live armed-sentinels are not orphaned on upgrade (regression test added).
- **P0-B** — `reload_lock._is_process_alive` now returns **True** on `PermissionError` (byte-parity with `spawn_lock._is_process_alive`): a live cross-user process must not be treated as dead and have its lock unlinked.
- **P0-C / M-1** — removed the redundant `[:12]` double-truncation at **both** sites that re-slice an already-≤12-char `_slug_for` result (`_reload_sentinel_path_for` and the bash-watcher writer in `_spawn_reload_watcher`) — they must byte-match or the daemon can't find the sentinel the watcher wrote if `_slug_for` is ever widened.
- **P0-D** — corrected a stale char-class comment in `spawn_lock`.
- **P0-E** — the parity tests now call the **real** `_reload_armed_path`/`_reload_ledger_path` rather than re-implementing the formula, so they can't stay green if the guard functions drift.
- **M-2** — restored the `str()` coercion the old inline formula provided, so a non-str `session_id` is handled defensively at the call site (callers are typed `str|None`; parity kept).

stdlib only; full suite green (pre-existing env-isolation flakes excluded).

### DIFF
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -2232,7 +2232,7 @@ def _spawn_reload_watcher(claude_pid: int, project_dir: str, session_id: str | N
     # The slug uses reload_lock._slug_for so it matches _reload_sentinel_path_for.
     from .reload_lock import _slug_for as _rl_slug_for
     if session_id:
-        sid12 = _rl_slug_for(session_id)[:12]
+        sid12 = _rl_slug_for(session_id)
         sentinel_path = f"/tmp/cozempic_reload_{sid12}.in-flight"
         status_path = f"/tmp/cozempic_reload_{sid12}.status"
         pgrep_pattern = f"claude.*{sid12}"
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -2755,8 +2755,9 @@ def safe_to_reload(team_state, messages, session_path) -> tuple[bool, str]:
 # precondition). `cozempic reload` clears the sentinel (user took control).
 # Shared by guard (write) and cli `nudge` (read/warn).
 def _reload_armed_path(session_id: str | None, session_path: Path | None = None) -> Path:
-    raw = (session_id or (session_path.stem if session_path else None) or "session")
-    slug = re.sub(r"[^a-z0-9_-]", "_", str(raw).lower())[:12] or "session"
+    from .reload_lock import _slug_for as _rl_slug_for  # lazy — matches guard.py:2233 pattern
+    raw = session_id or (session_path.stem if session_path else None) or None
+    slug = _rl_slug_for(raw) if raw else "default"
     return _guard_tmp_root() / f"cozempic_reload_armed_{slug}.json"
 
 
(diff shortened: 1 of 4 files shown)


ynaamane:
**`/code-review max` hardening pass pushed** (4 follow-up commits — see the updated description).

Key fixes: `guard._reload_armed_path`/`_reload_ledger_path` now call the canonical `reload_lock._slug_for` instead of inlining the formula (true cross-producer parity, not two copies that merely agree for UUIDs); `_is_process_alive` returns `True` on `PermissionError` (parity with `spawn_lock` — a live cross-user process must not have its lock unlinked); dropped a redundant `[:12]` at **both** the sentinel-path reader and the bash-watcher writer (they must byte-match); restored a defensive `str()` coercion at the call sites.

Adversarial review: PASS WITH CONDITIONS — both MED conditions folded in round-2. UUID slug-invariance preserved (no live armed-sentinel orphaning on upgrade). stdlib only.

PR #136, opened 2026-06-15
fix(guard): SIGTERM-leak + test /tmp-log isolation + _pid_is_alive consolidation (GC-1/2/3)

ynaamane:
Three guard.py lifecycle / resource cleanups (LOW–MED), none touching the hot reload path. Per-line analysis confirmed no overlap with the in-flight reload-gate PRs.

## GC-1 — SIGTERM startup-leak

`_graceful_shutdown` (the SIGTERM handler, registered *before* the daemon's main `try/finally`) called `checkpoint_team` + `sys.exit(0)` but **not** the pidfile / armed-file cleanup. If SIGTERM fired in the ~200-line startup window before the `try:`, the `finally` that unlinks the pidfile never ran → **pidfile AND armed-file leaked**.

Fix: extracted the handler to a testable `_make_sigterm_handler(session_id, session_path, overflow_watcher)` factory that calls `_safe_unlink_session_pidfile(session_id)` + `clear_armed(session_id, session_path)` before `sys.exit(0)`. The call-site passes `sess["session_id"]` (the *resolved* id — the bare CLI arg is `None` on the auto-detect path).

## GC-2 — test `/tmp` log leaks

Two tests wrote real `/tmp` log files with no cleanup (`test_guard_robustness.py`, `test_guard_reload_watcher_poll.py`). Fix: patch `_guard_tmp_root` → `tmp_path` / use `_guard_tmp_root()` for the log path, matching the production log location.

> Out of scope, flagged for follow-up: production's reload **status** file (guard.py:2262, documented GAP-B) still uses hardcoded `/tmp`, while the log uses `_guard_tmp_root()`. The test correctly checks the real `/tmp` status path; the production status-vs-log path inconsistency is a separate consistency item.

## GC-3 — `_pid_is_alive` consolidation

Three liveness implementations (`guard.py`, `session.py`, `watchdog.py`) → one canonical `_pid_is_alive` in `helpers.py` (a leaf module — no import cycle). `session.py` + `watchdog.py` now import it (identical function object). **Behavioral fix:** `session.py`'s old impl returned `False` on a POSIX unknown `OSError`; the canonical returns `True` (fail-open — the process most likely exists). `reload_lock` / `spawn_lock` `_is_process_alive` are deliberately **not** consolidated — they carry a purpose-specific PermissionError semantic (lock-acquisition liveness), distinct from general liveness.

## Tests

`tests/test_guard_cleanup.py` (new, 14 tests) — RED-at-base proven per item: GC-1 (AttributeError at base → handler-calls-cleanups after), GC-2 (`assertNotIn` hardcoded `/tmp` at base), GC-3 (ImportError at base → single-source `assertIs` + a POSIX-`OSError` fail-open behavioral test). Full suite: **1383 passed, 0 failed**.

## Scope

L3 / L4 resource + test isolation. `guard.py`, `session.py`, `watchdog.py`, `helpers.py`, 2 test files. stdlib only.

### DIFF
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -3641,19 +3641,18 @@ def _pid_identity_match(pid: int, session_id: str | None) -> bool:
     return abs(current_start_time - recorded_start_time) < 0.1
 
 
-def _pid_is_alive(pid: int) -> bool:
-    """Bare process-liveness probe — does NOT consult the JSONL mtime.
-
-    Anti-resurrection: a dead PID must read as dead even when cozempic's own
-    ``save_messages`` just refreshed the session JSONL moments earlier.
-    ``_is_claude_process``'s mtime fallback would misread that fresh write as a
-    live Claude and let the reload watcher resurrect a session the user closed.
-    ``os.kill(pid, 0)`` answers liveness directly and is not fooled by it.
-
-    Canonical implementation lives in helpers._pid_is_alive (GC-3);
-    this is the same function re-exported under the guard-internal name.
-    """
-    return _pid_is_alive_canonical(pid)
+# Bare process-liveness probe — does NOT consult the JSONL mtime.
+#
+# Anti-resurrection: a dead PID must read as dead even when cozempic's own
+# ``save_messages`` just refreshed the session JSONL moments earlier.
+# ``_is_claude_process``'s mtime fallback would misread that fresh write as a
+# live Claude and let the reload watcher resurrect a session the user closed.
+# ``os.kill(pid, 0)`` answers liveness directly and is not fooled by it.
+#
+# Canonical implementation lives in helpers._pid_is_alive (GC-3);
+# module-level alias so callers in this module and tests that patch
+# ``guard._pid_is_alive`` continue to work without a wrapper call.
+_pid_is_alive = _pid_is_alive_canonical
 
 
 def _is_claude_process(pid: int, session_path: Path | None = None) -> bool:
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -483,6 +483,33 @@ def _validate_finite_thresholds(
             raise ConfigError(f"{_name} must be a finite number, got {_v!r}")
 
 
+def _make_sigterm_handler(session_id, session_path, overflow_watcher):
+    """Return the SIGTERM handler for a guard daemon instance.
+
+    Extracted so it can be tested independently (GC-1). The handler:
+    1. Writes a final team checkpoint.
+    2. Stops the overflow watcher if one is running.
+    3. Unlinks the session PID file (CAS — only if it holds OUR pid).
+    4. Clears the armed-reload sentinel so a SIGTERM doesn't leave a stale
+       armed file that would cause the next daemon to reload immediately.
+
+    Steps 3+4 were missing before GC-1: a SIGTERM before the try: block at
+    ~896 in start_guard would exit without cleanup, leaving a leaked PID file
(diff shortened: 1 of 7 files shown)


ynaamane:
Converting to **draft — do not merge yet.** An adversarial review caught a CRITICAL regression in GC-3 that the test suite masks:

The consolidated `helpers._pid_is_alive` was taken from guard's int-only variant (`isinstance(pid, int)` → rejects non-ints), but it dropped the string coercion the old `session._pid_alive` had (`os.kill(int(pid), 0)`). `record_active_transcript`'s prune (`session.py:581`) passes dict **keys**, which are JSON strings (`"12345"`) — so `_pid_is_alive("12345")` now returns `False`, and the prune reads every entry as dead and drops live ones. This breaks the #124 two-terminal active-transcript store (the second terminal evicts the first).

The suite stayed green because every active-transcript test mocks `_pid_alive`, so the real string-key path is never exercised — and one new test enshrined the int-only contract.

Pushing a fix (coerce `int(pid)` in the helper + an **unmocked** `record_active_transcript` regression test with two live pids) plus a MED fix (wrap `checkpoint_team` in the SIGTERM handler so the pidfile/armed cleanup runs even if checkpoint raises). Will mark ready once verified.

ynaamane:
Added 4 robustness fixes from an independent code-review max pass over this PR (commits `b7e6615` test + `9380c29` fix + `ab3acea`):

- **A (SIGTERM handler):** the `finally` cleanups (`overflow_watcher.stop()`, `_safe_unlink_session_pidfile`, `clear_armed`) were unguarded and `sys.exit(0)` sat outside the `try/finally` — a raising `overflow_watcher.stop()` would skip the remaining cleanups AND `sys.exit(0)`, leaving the daemon alive after SIGTERM. Each cleanup is now individually `try/except`'d so `sys.exit(0)` always fires. RED-at-prev proven with a raising watcher — the `overflow_watcher != None` path was previously untested (both tests passed `None`).
- **C (`_pid_is_alive` coercion):** didn't catch `OverflowError`, so `float('inf')` raised instead of returning `False` (the old `session._pid_alive` caught it). Added `OverflowError`.
- **E (test leak):** the watcher-poll test's `guard_log` (`_guard_tmp_root()` = /tmp on POSIX) had no `addCleanup` → real /tmp leak. Added.
- **alias:** replaced the `guard._pid_is_alive` one-line wrapper with a module-level alias.

Full suite green (1423 passed, 0 failed).

ynaamane:
Marking ready — independently re-reviewed → **PASS** after the 4 robustness fixes (SIGTERM per-step cleanup hardening with the previously-untested `overflow_watcher`-raise path now covered + RED-at-prev proven; `OverflowError` coercion; test-leak `addCleanup`; module-level alias verified patchable — 40 monkeypatch tests pass). The earlier CRITICAL (the string-pid coercion regression in the `_pid_is_alive` consolidation) is closed with an **unmocked** `record_active_transcript` regression test (2 live pids, RED at the pre-fix commit). Full suite green (1423 passed / 0 failed). Ready for merge.

junaidtitan:
Thanks @ynaamane — merged to main (b1747c2), rebased onto current 1.8.33 (the branch base predated #138). All of GC-1/2/3 landed: the SIGTERM-handler factory wiring, the /tmp test-log isolation, and the `_pid_is_alive` consolidation into helpers. The only conflict was guard.py's helpers import line — resolved by keeping BOTH `hashable_str` (added on main via #138) and your new `_pid_is_alive` import. Full suite green. Credited to you in the commit. 🙏

PR #137, opened 2026-06-15
fix(guard): bound reload-ledger env knobs so the storm-guard can't be silently disabled (L1)

ynaamane:
## Problem — the reload-storm guard can be silently disabled

The reload-**storm** guard (which stops a runaway reload→regrow→reload loop) could be silently defeated four ways:

1. **Unbounded env knobs.** \`_reload_ledger_window_s\` / \`_reload_ledger_max\` read \`COZEMPIC_RELOAD_WINDOW_S\` / \`COZEMPIC_RELOAD_MAX\` with a floor but **no upper bound** — a huge value (\`COZEMPIC_RELOAD_MAX=999999999\`) makes the threshold astronomically large → it never trips.
2. **Incoherent ceiling vs write-cap.** Even the original ceiling of \`100\` was incoherent with the \`hist[-50:]\` write-cap: the ledger never holds >50 entries, so any \`RELOAD_MAX ∈ [51, 100]\` made \`len(hist) >= max\` **permanently False** — the guard was silently inert for that whole range.
3. **Large-finite gate-disable class.** \`_reload_warn_grace\` capped NaN/inf but not huge-finite values: \`COZEMPIC_RELOAD_WARN_GRACE=99999999999\` passes \`math.isfinite()\` but makes \`elapsed >= grace\` permanently False — the idle-reload fallback never fires.
4. **Non-atomic ledger write.** A bare \`ledger_path.write_text(…)\` can produce a half-written ledger on SIGKILL mid-write, which \`_reload_rate_exceeded\` would silently reset to \`[]\` — erasing the storm-guard counter and allowing unlimited reloads after a crash.

A removed-behavior regression was also folded in: the \`ae85bcc\` base commit silently dropped the whitespace-only-env warning in \`parse_env_positive_int\` / \`parse_env_non_negative_int\` (an undocumented behavior change).

## Fix

**P0-A** (\`guard.py\` \`_reload_warn_grace\`): add \`or v > 3600\` to the isfinite reject condition. Cap = 3600 s (1 h): a grace period above 1 h is functionally infinite for any interactive session. Preserve the \`≤0\` disable semantic.

**P0-B+C** (\`guard.py\` \`_reload_ledger_max\`): extract \`_RELOAD_LEDGER_HIST_CAP = 50\` (a named constant shared by the \`maximum=\` argument and the \`hist[-cap:]\` write-cap slice), lower the effective ceiling from 100 to 50, and remove the now-dead \`max(1, v)\` (parse_env_positive_int guarantees ≥1 when non-None). An explicit INVARIANT comment documents the coupling: a ceiling above the write-cap silently disables the guard for the (write-cap, ceiling] range.

**Atomic write** (\`guard.py\` \`_reload_rate_exceeded\`): replace bare \`write_text\` with the inline \`tmp + os.replace + finally-cleanup\` pattern (mirrors \`_write_armed_atomic\` at guard.py:2780). SIGKILL between the tmp write and the rename corrupts only the tmp — the live ledger is untouched and the finally block unlinks the orphan.

**P0-D** (\`tests/test_input_coercion_corpus.py\`): add rows for \`_reload_ledger_window_s\` / COZEMPIC_RELOAD_WINDOW_S and \`_reload_ledger_max\` / COZEMPIC_RELOAD_MAX to the \`_ENV_VALIDATORS\` table (the qa-fleet L1 maintenance contract: every knob must have a corpus row).

**Q-A** (\`_validation.py\`): restore whitespace-only-env warning in \`parse_env_positive_int\` and \`parse_env_non_negative_int\`. The \`ae85bcc\` base used \`raw.strip() == ""\` as the early-return guard, silently treating whitespace-only env vars as absent; origin/main warned. Restored with \`if raw:\` guard that keeps genuine empty string silent.

**Round-2 (reviewer findings):**
- **M-1**: Added \`test_crash_safety_old_ledger_preserved_on_failed_replace\` — patches \`cozempic.guard.os.replace → OSError\`, asserts (a) pre-existing ledger byte-intact and (b) no \`.tmp*\` orphan. RED at base ae85bcc (bare \`write_text\` never calls \`os.replace\` → patch inert → old content overwritten). GREEN at HEAD (atomic path raises, live file untouched, finally cleans tmp).
- **L-1**: Updated \`parse_env_bool\` docstring to document the deliberate whitespace-only divergence from \`parse_env_positive_int\`.
- **L-2**: Expanded \`_RELOAD_LEDGER_HIST_CAP\` inline comment into an explicit INVARIANT block.

**WONTFIX (documented):**
- L-3 (SIGKILL \`.tmp\` orphan not GC'd): inherited from the \`_write_armed_atomic\` sister pattern — low rate, no corruption, no fix needed.

## Tests

**New test classes / updates (7 commits):**
- \`TestReloadWarnGraceBounds\` (9 tests): huge-finite rejected, 3601 rejected, 3600 accepted, 0/negative disable-semantic preserved, nan/inf rejected.
- \`TestReloadLedgerMaxEnvBounds\`: corrected clamp→reject semantics; ceiling 100→50; \`test_max_above_write_cap_rejected_to_default\` (RELOAD_MAX=80 now rejected); \`test_ceiling_100_rejected_after_fix\`; \`test_zero_rejected_to_default\` (honest docstring).
- \`TestReloadLedgerWindowSEnvBounds\`: \`test_floor_clamp_at_60\` upgraded to \`assertEqual\` (not \`assertGreaterEqual\`).
- \`TestReloadLedgerAtomicWrite\` (3 tests): crash-safety (patch os.replace→OSError), no-orphan on clean write, valid JSON after write.
- \`TestParseEnvPositiveInt.test_warns_on_whitespace_only\`, \`TestParseEnvNonNegativeInt.test_warns_on_whitespace_only\` — Q-A regression guard.
- \`_ENV_VALIDATORS\` corpus: 7 → 9 rows.

**Suite:** 1400 passed, 5 skipped, 271 subtests, 0 failures (baseline 1383/233 on origin/main).

**RED-at-base proof (7 behavioral tests):** Each new bug-capture test was verified RED at ae85bcc (the PR branch base) and GREEN at HEAD. The M-1 atomicity test was independently reproduced by the lead.

## Scope

- \`src/cozempic/guard.py\` — \`_reload_warn_grace\`, \`_RELOAD_LEDGER_HIST_CAP\`, \`_reload_ledger_max\`, \`_reload_rate_exceeded\`
- \`src/cozempic/_validation.py\` — \`parse_env_positive_int\`, \`parse_env_non_negative_int\`, \`parse_env_bool\` (docstring)
- \`tests/test_reload_ledger_env_bounds.py\` — new test file
- \`tests/test_input_coercion_corpus.py\` — corpus extension
- \`tests/test_validation.py\` — Q-A regression guards

No changes to the public API, CLI, or PyPI packaging. Zero new dependencies.

### DIFF
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -267,8 +267,13 @@ def parse_env_bool(name: str, default: bool = False, warn: bool = True) -> bool:
     returns `default`.  Empty / absent / whitespace-only: returns `default`
     silently.
 
-    Follows the warn+fallback contract of parse_env_positive_int:
-    env vars are ambient config; an unrecognized value must not crash.
+    Follows the warn+fallback contract of parse_env_positive_int for
+    unrecognized non-empty values: emits a warning and returns `default`.
+    Deliberate divergence: whitespace-only input is treated as absent and
+    returns `default` silently (no warning), unlike parse_env_positive_int
+    which warns on whitespace-only.  Bool knobs have no numeric parse step
+    where whitespace ambiguity is meaningful, so the silent-absent treatment
+    is more appropriate.
 
     The `warn` parameter suppresses the warning when set to False — useful
     for in-body re-reads that run on every call (e.g. _debug()) where a
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -188,10 +188,10 @@ def parse_env_positive_int(name: str, *, maximum: int | None = None) -> int | No
     ``pct = total / window`` rounds toward 0 when window is huge).
     """
     raw = os.environ.get(name)
-    if raw is None or raw == "":
+    if raw is None or raw.strip() == "":
         return None
     try:
-        value = int(raw)
+        value = int(raw.strip())
     except ValueError:
         _env_warn(name, raw, "must be an integer")
         return None
@@ -217,10 +217,10 @@ def parse_env_non_negative_int(name: str, *, maximum: int | None = None) -> int 
     context window would zero out usable context — cap at the default window.
     """
     raw = os.environ.get(name)
-    if raw is None or raw == "":
+    if raw is None or raw.strip() == "":
         return None
     try:
-        value = int(raw)
+        value = int(raw.strip())
     except ValueError:
         _env_warn(name, raw, "must be an integer")
         return None
--- a/src/cozempic/_validation.py
+++ b/src/cozempic/_validation.py
@@ -188,10 +188,20 @@ def parse_env_positive_int(name: str, *, maximum: int | None = None) -> int | No
     ``pct = total / window`` rounds toward 0 when window is huge).
     """
     raw = os.environ.get(name)
-    if raw is None or raw.strip() == "":
+    if raw is None:
+        return None
+    stripped = raw.strip()
+    if stripped == "":
+        # Whitespace-only (raw non-empty) is set but carries no integer — warn
+        # so the user knows their override was ignored.  Genuine empty string
+        # (export VAR=) is unset-equivalent; silently ignore it.
+        # (Regression fix: ae85bcc dropped whitespace-only silently; origin/main
+        # let int(raw) raise ValueError → same "must be an integer" warn path.)
+        if raw:  # whitespace-only; raw=="" is the silent empty-string case
+            _env_warn(name, raw, "must be an integer")
         return None
     try:
-        value = int(raw.strip())
+        value = int(stripped)
     except ValueError:
         _env_warn(name, raw, "must be an integer")
         return None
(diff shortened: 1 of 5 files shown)


ynaamane:
**`/code-review max` hardening pass pushed** (see the updated description) — reframed as *storm-guard integrity*.

Key fixes: folded the large-finite gate-disable cap into `_reload_warn_grace` (class-of-bug fold of the `WINDOW_S`/`MAX` fix); made the `RELOAD_MAX` ceiling coherent with the `hist[-50:]` write-cap via a shared `_RELOAD_LEDGER_HIST_CAP` constant (values in [51,100] were silently ineffective — the guard could never fire); added an **atomic ledger write** (tmp + `os.replace` + `finally`-cleanup) with a RED-at-base crash-safety test (a SIGKILL mid-write no longer resets the storm history); restored a whitespace-only-env warning the original refactor had dropped.

Adversarial review: PASS WITH CONDITIONS — the atomicity-test condition folded in round-2. Full suite 1400 green.

PR #138, opened 2026-06-15
Audit-P1 remediation: merge #128/#129/#130 + close all whole-codebase-audit P1s

junaidtitan:
## Audit-P1 remediation: merge #128/#129/#130 + close all whole-codebase-audit P1s

Closes the confirmed P1s from the 2026-06-15 whole-codebase QA-fleet audit, and merges three clean contributor PRs, on one branch. **Suite green (1439 passed; 3 known env-flaky deselected). No release in this PR — landing only.**

### Merged contributor PRs
- **#130** — `enforce_floor` must not fork the DAG on a compacted session (closes the prune-safety P1)
- **#129** — skip sub-agent sidechain turns in `extract_corrections`
- **#128** — verify guard identity before `guard-watchdog --fix` SIGTERM (the *detection* half — the K/M regex — is fixed here too, so the watchdog P1 is closed in full)

### Audit P1 fixes
- **Data-loss**: `_FileSnapshot` equal-size re-hash; `tool-result-age` diff-collapse only inside a hunk; doctor `fix_corrupted_tool_use` snapshot + append-merge
- **TOCTOU (read-once)**: `_FileSnapshot.from_bytes` + `load_messages_and_snapshot` (snapshot↔load window) and `classify_and_delta` (classify↔read_delta window), routed through all 6 mutate-then-save callers
- **Unicode-separator corruption**: `_split_physical_lines` replaces `str.splitlines()` at all 4 raw-JSONL parse sites (U+2028/U+2029/U+0085 are legal in JSON strings and emitted unescaped by JS `JSON.stringify`)
- **Daemon crash**: `_as_int` usage coercion + per-cycle loop-body guard
- **Watchdog**: K/M/comma-suffixed prune-line regex (+ #128 identity gate)
- **Injection**: digest `_sanitize_for_injection` at every Claude-memory sink incl. `cmd_remind`
- **Windows**: PowerShell-native reload-watcher; fail-closed ReDoS budget when no SIGALRM
- **Recovery/quality**: overflow token-axis preflight + widened markers; doctor honest "fixed" (re-check); atomic-write consolidation to one hardened helper; MCP `treat_session` static→behavioral test

### Verification
6 adversarial QA-fleet gates run to convergence (P1 count 2→3→1→[1+1 P2]→0); each fix has a regression test; full suite after every fix.

### Confidence notes for the reviewer
- **Windows** fixes are verified by emulation only (no Windows CI runner) — a `windows-latest` job would close that.
- **overflow-marker** widened safely (strictly broader, can't regress) but full confidence needs a captured **real** overflowed-CC JSONL tail to pin the persisted text — flagged in-code.
- A few **P2/P3** nits remain as deliberate fast-follow.
- Out of scope: #131–137 (further @ynaamane PRs).

### DIFF
--- a/RELEASE_DRAFT_1.8.33.md
+++ b/RELEASE_DRAFT_1.8.33.md
@@ -1,0 +1,48 @@
+# Cozempic v1.8.33 — DRAFT (unreleased)
+
+> Status: on branch `remediation/audit-p1`, PR #138 (open, MERGEABLE). **Landing only — not yet released to any channel.** Verified across 6 adversarial QA-fleet gates run to convergence; full test suite green (1439 passed; 3 known env-flaky deselected).
+
+Whole-codebase audit remediation: closes every P1 from the 2026-06-15 adversarial QA-fleet audit and folds in three contributor PRs.
+
+## 🔒 Data integrity (silent-data-loss fixes)
+- **Equal-size overwrite** — `_FileSnapshot.classify()` re-hashes on an equal-*size* file instead of assuming "unchanged", so an in-place rewrite by Claude of the same byte length is no longer silently clobbered on prune-save.
+- **Read-once snapshots (TOCTOU ×2)** — a line appended in the window between snapshot and load could be duplicated, and a concurrent rewrite between `classify` and `read_delta` could merge an unvalidated tail. New `_FileSnapshot.from_bytes` / `load_messages_and_snapshot` (single read) and `classify_and_delta` (single read) close both windows; routed through all 6 mutate-then-save call sites.
+- **Unicode-separator corruption** — JSONL parsing used `str.splitlines()`, which splits on U+2028/U+2029/U+0085 (legal inside JSON strings, emitted unescaped by JS `JSON.stringify`), tearing a valid line into broken fragments. New `_split_physical_lines` (splits on `\n`/`\r` only) applied to all 4 raw-JSONL parse paths.
+- **`tool-result-age`** — diff-collapse only collapses context *inside* a real unified-diff hunk, so indented non-diff tool output (git-log, CI, config) is preserved verbatim.
+- **doctor `fix_corrupted_tool_use`** — snapshots + append-merges instead of clobbering, so a turn appended mid-repair is preserved.
+
+## 🛡️ Daemon resilience
+- **Crash on malformed `usage`** — a null/string token value no longer crashes the guard; `_as_int` coercion + a per-cycle loop-body guard keep one bad cycle from killing the daemon.
+- **Guard-loop watchdog** — fixed the K/M/comma-suffixed prune-line regex so a healthy daemon is no longer false-flagged as looping; with the merged #128 identity gate, `guard-watchdog --fix` can no longer SIGTERM a productive guard.
+
+## 🪟 Windows
+- **Auto-resume** — replaced the dead cmd-in-bash reload-watcher with a PowerShell-native watcher.
+- **ReDoS budget** — the `--protect-pattern` match budget (POSIX-only SIGALRM) now fails *closed* on Windows: a catastrophic-backtracking pattern is refused up front instead of freezing the daemon.
+
+## 🧹 Security & correctness
+- **Digest prompt-injection** — untrusted rule/evidence text is sanitized (single-line, markdown-defanged) at every sink that writes into Claude's memory, including the `cmd_remind` hook.
+- **Overflow recovery** — added a post-prune *token*-axis preflight (was byte-only) and widened the overflow-marker set.
+- **doctor `--fix`** — reports "fixed" only when a re-check confirms the issue is gone (no more false success on a no-op/failed fix).
+
+## 📦 Merged contributor PRs (@ynaamane)
+- **#130** — `enforce_floor` no longer forks the conversation DAG on a compacted session
+- **#129** — sub-agent (sidechain) turns excluded from learned corrections
+- **#128** — guard-identity verification before a `--fix` SIGTERM
+
+## 🧰 Internal
+- Consolidated three atomic-write implementations into one hardened helper.
+- Replaced false-confidence static tests (MCP `treat_session`, watchdog) with behavioral ones.
+- +60-odd regression tests across the touched subsystems.
+
+---
+
+## Confidence ledger (for reviewers)
+- **Full confidence** (offline-verified + regression-tested): everything except the two below.
+- **Emulation-only**: the two Windows fixes — no Windows CI runner; verified by emulating the no-SIGALRM / `os.name == 'nt'` paths.
+- **Artifact-blocked**: the overflow-marker widening is strictly broader (cannot regress detection) but full confidence needs a captured **real** overflowed-CC JSONL tail to pin the persisted marker text (flagged in-code).
+
+## Still pending
+**To land:** merge PR #138; cut the release (PyPI / npm / GitHub / Homebrew tap / packaging mirrors) as a separate deliberate step.
+**To lift the caveats:** add a `windows-latest` CI job; capture a real overflowed-CC JSONL fixture.
+**Fast-follow:** a few P2/P3 nits ranked below the ship line; move `packaging/ci/publish.yml` (Trusted Publishing) into `.github/workflows/` once a `workflow`-scoped token is available.
+**Out of scope this run:** PRs #131–137; the de-scoped npm `COZEMPIC_NO_GLOBAL_INIT` opt-out; the team-checkpoint→post-compact untrusted-text surface (distinct pre-existing class).
--- a/plugin/servers/cozempic_mcp.py
+++ b/plugin/servers/cozempic_mcp.py
@@ -142,7 +142,7 @@ def treat_session(prescription: str = "standard", execute: bool = False) -> str:
     """
     from cozempic.session import (
         _PruneLock, PruneConflictError, PruneLockError,
-        find_current_session, load_messages, save_messages, snapshot_session,
+        find_current_session, load_messages, load_messages_and_snapshot, save_messages,
     )
     from cozempic.registry import PRESCRIPTIONS
     from cozempic.executor import run_prescription
@@ -159,12 +159,11 @@ def treat_session(prescription: str = "standard", execute: bool = False) -> str:
         return f"Unknown prescription '{prescription}'. Options: {', '.join(PRESCRIPTIONS)}"
 
     path = sess["path"]
-    # Take snapshot BEFORE load so append-conflict detection works (parallel
-    # to cmd_treat/cmd_strategy/cmd_reload in cli.py). Without this, a guard
-    # daemon prune cycle running concurrently could silently overwrite our
-    # output — the same data-loss path Wave 1 fixed in the CLI.
-    snapshot = snapshot_session(path) if execute else None
-    messages = load_messages(path)
+    # Read once: snapshot + messages from identical bytes for append-conflict
+    # detection (parallel to cmd_treat/cmd_strategy/cmd_reload in cli.py), without
+    # a TOCTOU window where a concurrent append lands in both messages and delta.
+    messages, _snap = load_messages_and_snapshot(path)
+    snapshot = _snap if execute else None
     strategy_names = PRESCRIPTIONS[prescription]
 
     original_bytes = sum(b for _, _, b in messages)
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -979,17 +979,16 @@ def cmd_guard_watchdog(args):
         print(f"      {h.report.reason}")
         print(f"      futile={h.report.futile_cycles} total={h.report.total_prune_cycles} "
               f"backoff_max={h.report.max_backoff_s}s exit_seen={h.report.has_exit}")
-        if getattr(args, "fix", False) and h.pid_alive and h.pid:
-            if not h.guard_confirmed:
-                print(f"      → pid {h.pid} is alive but is NOT a cozempic guard "
-                      f"(recycled?) — refusing to kill")
-            else:
-                try:
-                    os.kill(int(h.pid), signal.SIGTERM)
-                    print(f"      → SIGTERM sent to looping daemon pid {h.pid}")
-                    arrested += 1
-                except (ProcessLookupError, PermissionError, OSError) as e:
-                    print(f"      ! could not signal pid {h.pid}: {e}")
+        if getattr(args, "fix", False) and h.guard_confirmed:
+            try:
+                os.kill(int(h.pid), signal.SIGTERM)
+                print(f"      → SIGTERM sent to looping daemon pid {h.pid}")
+                arrested += 1
+            except (ProcessLookupError, PermissionError, OSError) as e:
+                print(f"      ! could not signal pid {h.pid}: {e}")
+        elif getattr(args, "fix", False) and h.pid_alive and h.pid:
+            print(f"      → pid {h.pid} is alive but is NOT a cozempic guard "
+                  f"(recycled?) — refusing to kill")
         elif h.pid_alive:
             print(f"      → re-run with --fix to terminate this looping daemon "
                   f"(or: kill {h.pid})")
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -23,7 +23,7 @@ from .init import run_init
 from .recap import save_recap
 from .registry import PRESCRIPTIONS, STRATEGIES
 from .safety import PruneValidationError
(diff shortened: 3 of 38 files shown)


junaidtitan:
@ynaamane — would you be up for a peer review on this? 🙏

It folds in your **#128 / #129 / #130** and closes the rest of the 2026-06-15 whole-codebase audit P1s on one branch (verified across 6 QA-fleet gates run to convergence; suite green at 1439). Full per-area summary in [`RELEASE_DRAFT_1.8.33.md`](RELEASE_DRAFT_1.8.33.md).

Areas where your eyes would be most valuable:
- **`session.py` read-once refactor** — `_FileSnapshot.from_bytes` + `load_messages_and_snapshot` (snapshot↔load TOCTOU) and `classify_and_delta` (classify↔read_delta TOCTOU), routed through all 6 mutate-then-save callers. This is adjacent to your #130 floor work.
- **`_split_physical_lines`** replacing `str.splitlines()` at all 4 raw-JSONL parse sites (U+2028/U+2029/U+0085 corruption).
- **watchdog K/M regex** sitting under your #128 identity gate — together they close the watchdog P1.

Two honest caveats flagged in the draft I'd love a second opinion on: the **Windows** fixes are emulation-verified only (no Windows CI runner), and the **overflow-marker** widening is artifact-blocked on a real captured overflowed-CC JSONL. No rush — landing only, no release attached to this PR.

junaidtitan:
## ⚠️ DO NOT MERGE — converted to draft

An extreme adversarial mission-critical fleet (166 agents, "assume the fixes are wrong and make a user lose data") returned **ship: false** — **0 catastrophic, 0 P0, but 16 confirmed P1** (3-skeptic majority-voted, with repros). **Several are behavior regressions introduced by this very branch.** Blocking until remediated. Root-cause clusters:

1. **Decode corruption (regression):** `load_messages_and_snapshot` decodes `errors='replace'`, so an invalid-UTF-8 byte (binary tool_result, truncated multibyte) is silently rewritten to U+FFFD and saved — the old strict `load_messages` safely *aborted*. Loader trifecta now disagrees byte-for-byte.
2. **Inert-guard (regression):** the new per-cycle `except Exception: continue` turns a *deterministic* per-cycle error into a permanently inert-but-alive daemon — invisible to the watchdog (emits no `Pruned:` line). Fed by `_as_int(float('inf'))` → OverflowError and a non-dict `message` crash in `quick_token_estimate`.
3. **`_as_int(inf)` OverflowError** — `int(float('inf'))` raises; `1e999`/`Infinity` are valid JSON.
4. **`gentle.py:256` incomplete fix** — `msg.get('message',{}).get('usage')` crashes when `message` is a non-dict, on the very line the PR hardened.
5. **Injection sibling missed:** `team.py` to_markdown / to_recovery_text embed untrusted `result_summary`/`lead_summary` into CC memory unsanitized — the literal sibling of the digest `_sanitize_for_injection` fix.
6. **Overflow false-kill (regression):** widened `OVERFLOW_MARKERS` now match substrings users type in normal discussion → unsolicited kill+resume, and the reactive kill path lacks an `agents_active`/`safe_to_reload` gate.
7. **Watchdog blind spots:** false-negative on no-`Pruned:` reload-loops; log-injectable via attacker-controlled `team_name`.
8. **Windows ReDoS gate incomplete:** `_REDOS_SHAPE` misses alternation-overlap `(a|a)+`, brace-quantifiers, optional-chains, alternation-of-groups (emulation-only, untested on real Windows).
9. **Memory amplification:** read-once + `_split_physical_lines` → ~2-4× transient peak on large sessions (ungated on cli/mcp paths).

Remediation in progress on the branch; will re-run the mission-critical fleet on the fixes before this leaves draft.

ynaamane:
## Independent max-recall peer review — verified against head `ecf100f6` (r11)

Ran an independent review over the branch (parallel finder angles → adversarial verify → repro per finding, all against the current head). **First, the good news:** I could confirm a lot of the remediation landed solidly — the **C4 non-dict `message` root guard** in `_parse_one_line` (the altitude fix — wraps a non-dict inner `message` at the single loader chokepoint, so no consumer sees it), the **`errors="surrogateescape"`** sweep across the read paths, the **`_pid_alive`** consolidation, and the whole **r8→r11 ReDoS iteration** (r11 counting `?` as a branch quantifier closes the optional-chain `a?a?…` case the count rule missed). 👍

Below are the gaps I could still reproduce on top of that. Severities are my read against the daemon's own asymmetry (over-defer = recoverable; wrongly killing live Claude work = catastrophic); repros available on request.

### HIGH

**1. `overflow.py:347` — the reactive recovery's only in-flight-work safety gate FAILS OPEN.**
```python
try:
    _safe, _reason = _safe_to_reload(_extract_team_state(_msgs), _msgs, self.session_path)
except Exception:
    _safe, _reason = True, ""   # gate must never itself block a needed recovery
```
This is the only thing protecting a live agent team on the reactive path. Any throw in `load_messages` / `extract_team_state` / `safe_to_reload` → `_safe=True` → it proceeds to kill+resume (SIGKILL) a session with running subagents. That **inverts the asymmetry the rest of the daemon is built on** — on a gate error the safe move is to *defer* (the prune is already saved; Claude resumes from the pruned file next cycle), not to kill. And the trigger is real: `extract_team_state` still crashes on a malformed tool_result (see #2) — `TypeError: sequence item 0: expected str instance, int found`, which this gate swallows as "safe to kill." Fix: fail-CLOSED (`_safe, _reason = False, "safety gate error"`).

### MED

**2. `team.py:468` (+ `team.py:1014`) — a non-str `text` crashes `extract_team_state`.**
`_extract_block_text` does `"".join(sub.get("text","") for sub in content if … sub.get("type")=="text")` with no str-guard; `team.py:1014` does `block.get("text","")[:300]`. Reached via the tool_result path (829/847). Repro (head `ecf100f6`): a tool_result whose content is `[{"type":"text","text":99999}]` → `TypeError` out of `extract_team_state`. This is the exact respawn-storm class the PR hardens elsewhere, and it's *not* strategy-isolated, so its crash escapes into the guard cycle escalation **and** trips #1's fail-OPEN. Coerce `text` to `str`.

**3. `overflow.py:314/349/351` — the new 5b safe-point defer branch mis-accounts the circuit breaker.**
`self.breaker.record_recovery(...)` runs at step 5 (314) *before* the 5b gate, but the gate can `return` early (349-353) without any kill+resume. So (a) a deferral around legitimately-busy in-flight work still **consumes a breaker slot** → 3 benign defers trip `BREAKER_MAX_RECOVERIES` and silently disable the reactive net (the proactive path at `guard.py:1775-1780` deliberately does *not* count a safe-point defer — the reactive path breaks that sister-path invariant); (b) the log says `"prune saved, no resume this cycle"` but on this branch only `checkpoint_team` runs, which does **not** persist the prune (the pruned file is written only by `_write_pruned_after_exit`, inside `_terminate_and_resume`, post-kill) — so the prune is *not* saved; (c) no re-entry throttle, so every file-growth re-runs a full `guard_prune_cycle` until the breaker trips. Move `record_recovery` after the gate passes; fix the log to "prune not persisted (deferred)".

**4. `guard.py:1687` — `tokens_saved` re-introduces the `and post_te.total` trap you fixed 57 lines up.**
`tokens_saved = pre_te.total - post_te.total if pre_te.total and post_te.total else 0`. Your `_tokens_saved_now` at 1629-1630 correctly gates on `pre_te.total` only (with the NB comment at 1626-1627 naming this exact failure mode), but the `_record_persisted_savings` sibling here still has `and post_te.total` — a maximal prune to `post_te.total == 0` reads as 0, so the largest-possible savings event is never recorded. Same one-line fix.

**5. `team.py:1178` — `write_team_checkpoint` is non-atomic.**
`path.write_text(state.to_markdown(), …)` while `digest` was just consolidated onto `atomic_write_text` (and every other shared-state writer is atomic). A SIGKILL/OOM mid-write leaves a partial checkpoint that PostCompact reads back as recovery state. One-line swap to `atomic_write_text`.

**6. `executor.py:296` — the per-strategy `except Exception` swallows `ConfigError`.**
`strategy_*` funcs call `coerce_non_negative_int(config, …)` internally (e.g. `aggressive.py:261/333`), which raises `ConfigError(ValueError)` on a bad config value. The isolation `except Exception` catches it, prints "malformed message?", and skips the strategy — so a user's *config* error is silently hidden and mislabeled as a data problem, defeating the very `coerce_*` validation this PR adds. Re-raise `ConfigError` (or catch it separately with a config-specific message) before the generic `except`.

**7. `watchdog.py:152` — the new C7 escalation rule false-flags a healthy guard across daemon generations.**
`if rep.cycle_escalations >= 2 or …` is **count-only** (no dominance/dilution gate, unlike the futile branch), and `cycle_escalations` is counted over the whole 256 KB log tail (145). Guard logs are keyed by `md5(cwd)[:12]` and opened in **append mode** (`guard.py:3308/3387`), so escalation lines from prior *dead* generations accumulate in one file. A currently-healthy generation in a respawn-prone cwd inherits ≥2 stale `cycle-error escalation` lines → `looping=True` → a `watchdog --fix` SIGTERMs a healthy guard (and a plain report misleads the operator). Scope the escalation/error counts to the current generation (split the tail on the last `--- Guard daemon started ---` marker — `_DAEMON_START_RE` already exists for `daemon_starts`).

### LOW

- **`session.py:1026`** — `load_messages_incremental` is the one remaining `errors="replace"` decode (lossy U+FFFD) while every sibling moved to `surrogateescape`; read-only, so the only effect is `byte_len` accounting skew on non-UTF-8 lines, but it contradicts the PR's own "keep this path consistent with `load_messages()`" comment. One-line swap.
- **`helpers.py:606`** — r11 fixed the group-marker `?` in rule-1 (`_count_variable_quantifiers`) but not rule-2: `_scan_quantified_group_bodies` slices the body as `pattern[start+1:i]` without stripping the group marker, so `_body_has_inner_quantifier` sees the leading `?` of `(?:…)`, `(?i:…)`, `(?P<n>…)` and over-rejects safe non-capturing/flag/named groups (`(?:abc)+` flagged risky). Strip the `(?…)` marker before scanning the body.
- **`aggressive.py:106`, `standard.py:421/165`** — these iterate `get_content_blocks(msg)` and call `block.get(...)` with no `isinstance(block, dict)` guard, so a non-dict content element crashes the strategy. Contained by the executor's per-strategy isolation (so not a storm) but it violates the contract `get_content_blocks`' own docstring states ("consumers must isinstance-guard each element"); the strategy just silently does nothing on a poison block. One-line guard each.
- **`guard.py:2359`** — the new Windows PowerShell reload-watcher's Phase-4 success poll uses `Get-Process -Name claude | Select-Object -First 1` (any claude), whereas the POSIX path is session-scoped (`pgrep -f 'claude.*{sid12}'`, with a regression test). Impact is limited to a masked resume-failure diagnostic on a multi-session Windows host (the Phase-1 *kill* is PID-scoped, so no wrong kill), but it reintroduces the exact non-session-scoped poll the POSIX side was hardened against.
- **`overflow.py:197`** (plausible) — the reactive detector was narrowed to a structural gate (`isApiErrorMessage` / `type==error` / …), and the code's own comment flags the real overflowed-CC JSONL shape as unverified ("ARTIFACT-BLOCKED"). If the real persisted shape carries none of those keys, the reactive net is permanently inert for genuine overflows. It's a backup to the proactive 4-tier guard, so not an open hole — but worth a captured-artifact golden fixture pinning the field name.

Happy to open issues or small PRs for any of these — **#1 (`overflow.py:347` fail-OPEN)** is the one I'd prioritize, since a malformed tool_result (#2) is a real, reproducible path into it.

junaidtitan:
## Mission-critical adversarial hardening — converged, ship-clean ✅

This PR was put through **13 rounds** of an adversarial AI-agent fleet (assume-the-fix-is-wrong, poisoned-input fuzzing, real concurrency, 3-skeptic majority-vote verification, completeness critic, ship/no-ship). It targets the **live transcript write path + guard daemon for ~70k users**, so the bar was zero reachable catastrophic (P0/P1) bugs.

**Round 13 verdict: `ship:true` — 0 P0/P1, 0 P2.** Full suite **1474 passed**.

### What the rounds closed (root-level fixes)
- **Silent data-loss / corruption:** sentinel-key collision (a transcript carrying our internal `_raw`/`_parse_error` markers could overwrite real content); non-UTF-8 byte handling now lossless (`surrogateescape` + byte-exact `_jsonl_line`, never corrupts, never freezes the guard); strict no-op/equal-size rewrite guards.
- **Daemon death / respawn-storm:** systemic `hashable_str` + loader uuid/parentUuid wrap closed the entire "unhashable field used as a set/dict key" crash class across executor / safety / guard / doctor / strategies; per-strategy isolation; bounded escalation defer.
- **Non-dict robustness:** every consumer of untrusted JSONL shapes (tokens, team, doctor, recap, diagnosis) guarded; shared `get_dict_blocks` for read-only consumers.
- **Overflow false-fire:** reactive kill+resume now only triggers on a genuine API-error entry (no more false-kill on a truncated tail fragment).
- **CC-memory injection:** all untrusted team/digest fields sanitized before reaching Claude's memory.
- **ReDoS (no-SIGALRM path):** after proving precise detection un-perfectable, the detector now **categorically** refuses the three backtracking sources (≥2 variable quantifiers incl `?`, quantified-ambiguous group, any alternation `|`). Fail-open (warn + skip, content stays prunable), Windows/non-main-thread only; the 512-char cap is the load-bearing backstop.

### Documented residual (non-blocking, P3)
The ReDoS over-rejection skips benign multi-quantifier / alternation `--protect-pattern`s on the no-SIGALRM path only — **fail-open** (no freeze, no data loss); POSIX main-thread users are unaffected. The zero-over-rejection durable fix (killable-subprocess hard timeout) is deferred as a follow-up; `_NO_BUDGET_MATCH_CAP=512` must stay pinned.

Lifting from draft. Merge decision is the maintainer's.

junaidtitan:
@ynaamane — thank you, this was an excellent review. Every one of your findings was valid and reproduced on head; all are now fixed (head `d0ff322`), and each fix was put back through the adversarial re-gate (rounds 14–16), which twice caught a regression in *my own fix* before it could ship — exactly the value of a second set of eyes. Point by point:

**HIGH**
1. **`overflow.py` fail-OPEN gate** — fixed. The safe-point gate now FAILS CLOSED: any throw in `safe_to_reload`/`extract_team_state` DEFERS the kill instead of proceeding to SIGKILL a live team. (`+ a regression test that makes the gate throw and asserts no kill.`)

**MED**
2. **non-str `text` crashing `extract_team_state`** — fixed (`_extract_block_text` join + the `[:300]` slice are str-guarded). This was indeed the reachable trigger into #1.
3. **breaker mis-accounting on the safe-point defer** — fixed: `record_recovery` moved AFTER the gate (a benign defer no longer burns a slot), the log corrected ("prune NOT persisted this cycle"), and a re-entry cooldown added.
4. **`tokens_saved` `and post_te.total` trap** — fixed (gate on `pre_te.total` only).
5. **non-atomic `write_team_checkpoint`** — fixed (`atomic_write_text`). Note: the first cut used `errors="surrogatepass"`, which the re-gate caught as *asymmetric* (it moved a `UnicodeDecodeError` to `read_team_checkpoint`/PostCompact). Final fix sanitizes lone surrogates → `U+FFFD` in `_san` (the render chokepoint) so the checkpoint is clean UTF-8 for cozempic *and* CC, with a tolerant read for any stale file.
6. **`ConfigError` swallowed by per-strategy isolation** — fixed (re-raised before the generic `except`).
7. **watchdog cross-generation false-flag** — fixed, and your finding drove the most work: generation-scoping (my first attempt) *let an error-storm escape*, so the detector now flags `cycle_errors >= loop_trip AND cycle_errors > productive_prunes`. Two **report-only** residual edge cases remain on a flat append-log (a stale-prune FN and a scattered-recovered-error FP); since the watchdog is report-only (`--fix` is manual + identity-gated) and the daemon's own circuit-breaker self-arrests storms, we're shipping this and tracking the durable rate-based redesign as **#140**.

**LOW** — all fixed: `load_messages_incremental` → `surrogateescape`; rule-2 strips the `(?:`/`(?flags:`/`(?P<n>` group marker so `(?:abc)+` isn't over-rejected; `isinstance` guards in the named aggressive/standard sites; the Windows reload-watcher success poll is now session-scoped (`Get-CimInstance` command-line match). The artifact-blocked overflow-marker shape stays documented in-code pending a captured real overflowed-CC JSONL.

Re-gate verdict (round 16): **ship: true, 0 P0/P1 reachable through an automatic path.** Suite 1487 green. Thanks again — the daemon-asymmetry framing ("over-defer recoverable; wrongly killing live work catastrophic") was exactly the right lens.

PR #139, opened 2026-06-16
Review fixes for #138: fail-CLOSED reactive safe-point gate (HIGH) + 2 crash/savings siblings

### LINKED ISSUE #2: Please make it a Claude Code plugin
This way it would be so much easier to install and update it.

ynaamane:
Follow-up to my review comment on #138 (https://github.com/Ruya-AI/cozempic/pull/138#issuecomment-4719614747) — this targets the `remediation/audit-p1` branch with the 3 cleanest, independently-verified fixes from that review (the HIGH + two MED). RED→GREEN per fix; full suite green. Merge into #138 or cherry-pick as you prefer.

## 1. [HIGH] overflow.py — reactive safe-point gate fails OPEN
**Problem:** `OverflowRecovery._do_recover`'s 5b safe-point gate — the only in-flight-work protection on the reactive path — did `except Exception: _safe, _reason = True, ""`. Any throw in `load_messages`/`extract_team_state`/`safe_to_reload` therefore proceeds to kill+resume (SIGKILL) a session with a live agent team. That inverts the daemon's core asymmetry (over-deferring a reload is recoverable; wrongly killing live in-flight work is catastrophic), and the trigger is real: a malformed `tool_result` crashes `extract_team_state` (fix #2), which this gate swallowed as "safe to kill."
**Fix:** fail-CLOSED — on a gate error, defer the kill (the prune is already saved; the guard resumes next cycle).

## 2. [MED] team.py — a non-str `text` field crashes `extract_team_state`
**Problem:** `_extract_block_text` (`"".join(sub.get("text", "") …)`) and the lead-summary builder (`block.get("text", "")[:300]`) assume `text` is a `str`. A non-str value (int/None/list) in malformed JSONL raises `TypeError` out of `extract_team_state` — which runs OUTSIDE per-strategy isolation, so it escapes into the guard cycle (and trips #1's fail-OPEN).
**Fix:** `isinstance(t := …get("text"), str)`-guard both sites (skip non-str text).

## 3. [MED] guard.py — a maximal prune to 0 tokens is never recorded
**Problem:** `tokens_saved = pre_te.total - post_te.total if pre_te.total and post_te.total else 0`. A maximal prune to `post_te.total == 0` is full progress (`pre - 0`), but the falsy `and post_te.total` reads it as 0, so `_record_persisted_savings` early-returns at `if tokens_saved <= 0` and the largest-possible savings event is never recorded. This is the exact trap the sibling `_tokens_saved_now` already guards correctly (`if pre_te.total` only, with the NB comment).
**Fix:** extract `_persisted_tokens_saved(pre, post)` (gate on `pre` only) and route BOTH savings sites through it — the `_tokens_saved_now` sibling was still on the old falsy-trap form, so this also folds that instance (class-of-bug fold).

## Tests
New `tests/test_pr138_review_fixes.py` — RED→GREEN per fix, each proven RED at the base commit:
- **overflow:** at base the gate fails-OPEN → `_terminate_and_resume` fires (the SIGKILL); after the fix it defers and `checkpoint_team` runs. Includes a control test asserting the happy path (gate returns safe) still kills.
- **team:** at base `extract_team_state` raises real `TypeError` on int/None/list `text` (tool-result and lead-summary paths); after, no crash.
- **guard:** a call-site integration test drives the real `_deferred_writer` → `_record_persisted_savings` chain and asserts `record_savings` is called with the pre total for a `post == 0` prune — RED at base via a *wrong value* (`record_savings` not called), not ImportError; plus pure-helper unit coverage.

Full suite: **1453 passed, 5 skipped** (`pytest --ignore=tests/test_model_detection.py`, the 3 pre-existing model-detection failures on this branch's base excluded).

## Scope / not included
Deliberately focused on the 3 fixes above. The strategy-internal text / non-dict-block crash siblings flagged in the review comment (`strategies/standard.py` `strategy_tool_output_trim` / `strategy_dedup_reminders`, `aggressive.py` `strategy_error_retry_collapse`) are **out of scope for this PR**: they run INSIDE the executor's per-strategy isolation, so a crash is contained (that strategy is skipped — degraded prune savings, not a guard-cycle wedge or respawn storm) — a materially different blast radius than the `extract_team_state` path fixed here. Happy to add 1-line `isinstance(text, str)` guards for those as defense-in-depth in a separate change if useful.

### DIFF
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -142,6 +142,24 @@ def _read_min_prune_ratio() -> float:
 _MIN_PRUNE_RATIO = _read_min_prune_ratio()
 
 
+def _persisted_tokens_saved(pre_total: int, post_total: int) -> int:
+    """Return the number of tokens freed by a prune cycle.
+
+    Gate on *pre_total* only — a maximal prune to post_total==0 is FULL progress
+    (pre − 0), not zero.  The old inline expression ``pre - post if pre and post``
+    was falsy-trapped: ``and post_total`` evaluates to 0 when post==0, so the
+    entire ternary returned 0 and the largest possible savings event was never
+    recorded in the lifetime tracker.
+
+    NB: the sibling ``_tokens_saved_now`` expression at the futile-reload gate
+    already uses the correct single-gate form (``if pre_te.total``); this helper
+    makes both sites share the same semantics.
+    """
+    if not pre_total:
+        return 0
+    return pre_total - post_total
+
+
 def _hard_prune_counts_as_futile(result: dict) -> bool:
     """Whether a HARD-tier prune cycle counts toward the futile-loop K-exit counter.
 
@@ -1684,7 +1702,7 @@ def guard_prune_cycle(
     # counters on every deferred or looping cycle that never actually wrote — the
     # in-the-wild prune-spike signature: a stuck guard re-pruning every interval
     # bumps the counter each time despite persisting nothing.
-    tokens_saved = pre_te.total - post_te.total if pre_te.total and post_te.total else 0
+    tokens_saved = _persisted_tokens_saved(pre_te.total, post_te.total)
 
     def _record_persisted_savings():
         """Record savings to the lifetime tracker / global counter. Call ONLY
--- a/src/cozempic/guard.py
+++ b/src/cozempic/guard.py
@@ -1641,12 +1641,9 @@ def guard_prune_cycle(
             # ALLOWS a prune that frees real headroom to reload even if it lands
             # just above the trigger band — that converges over a cycle or two, and
             # the disk reload-rate ledger below bounds any residual regrow loop.
-            # NB: gate on pre_te.total only — a maximal prune to post_te.total==0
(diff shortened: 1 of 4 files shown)


junaidtitan:
Thanks @ynaamane. Reviewed this carefully (multi-agent fleet + adversarial verify). The three intended fixes here (fail-CLOSED reactive safe-point gate, the str-guards, and savings-on-maximal-prune) are all correct — but they already landed on `main` via #138 (1.8.33), so this branch is redundant against them. More importantly the branch is based on `remediation/audit-p1` from *before* #138 finalized, so a real merge (`git diff origin/main`) would REVERT live 1.8.33 hardening and downgrade the version across the packaging files — the clean-looking `gh pr diff` is measured against the stale base.

So rather than merge/rebase, I salvaged the one genuinely net-new piece — the `_persisted_tokens_saved` helper (your falsy-trap fix: a maximal prune to 0 was recorded as 0 savings) plus its tests — onto main in c1d52250, crediting you. Closing this as superseded; the value is live. 🙏

PR #146, opened 2026-06-23
fix(dashboard): bound v1.8.34 prune-metrics coercions against the L1 adversarial corpus

ynaamane:
## Problem

The v1.8.34 prune-metrics dashboard added new coercions that are not bounded against the qa-fleet **L1** adversarial corpus. A corrupt or tampered `~/.cozempic` receipt / lifetime ledger crashes `cozempic dashboard` (uncaught) or silently renders wrong values:

- `aggregate._context_pct`, `lifetime.load_lifetime`, `render._fmt_tokens`/`_fmt_bytes` raise **`OverflowError`** on a huge int → `cmd_dashboard` crashes (the `load_lifetime`/`aggregate` calls are unwrapped). `load_lifetime`'s own docstring claims *"It never raises."*
- `aggregate()` raises **`TypeError`** when a receipt's `ts` is a non-string (the timeline/per-session sort compares `int` vs `str`).
- `_context_pct` silently accepts `bool` (`True/True → 100.0`) and negative `context_window` (`→ -500000.0`).
- `validate_receipt` does not reject `NaN`/`inf`, and `serialize_receipt` used `allow_nan=True` → invalid `NaN`/`Infinity` JSON literals could be written.
- `receipts_enabled` (`not os.environ.get(...)`) treated `=0`/`=false` as opt-out — disabling receipts when the user meant to keep them on.
- `build_receipt`'s strategy apportionment used float division → `OverflowError` on a huge total zeroed **all** strategy attributions (including valid ones).

Observability-path only (no guard/prune/reload impact, no data loss), but the L1 class this project hardens proactively — and the new modules had **no** coercion-corpus coverage.

## Fix

- New shared cap `_MAX_RECEIPT_INT = 10**15` in `_constants.py` (far above any real token/byte count; keeps `huge / small * 100` finite).
- Bound + type-guard every coercion, mirroring the existing `_int` idiom: `_context_pct` (bool-exclude, `window > 0`, `after >= 0`, both `<= cap`, result capped at `100%`), `_int` / `_num_or_zero` (magnitude `> cap` and negatives → `0`, sibling-consistent; `load_lifetime` honors *"never raises"*), `_fmt_tokens`/`_fmt_bytes`/`_fmt_int` (clamp before float, sign-preserving), `session_multiplier_x` capped.
- `aggregate()` timeline/per-session **sort keys coerced to `str`** (a non-string `ts` can no longer crash the sort); strategy `id`/`tier` `or "unknown"` (explicit `null`).
- `build_receipt` strategy apportionment rewritten in **pure integer Hamilton (largest-remainder)** — `divmod` per strategy, leftover distributed by largest remainder. Overflow-immune for arbitrary Python ints; attribution preserved (no more all-zero on a huge total).
- `validate_receipt` rejects non-finite `tokens.*` / `bytes.*` / `model.context_window`; `serialize_receipt` and the receipt index use **`allow_nan=False`** (the production output-boundary guard — `validate_receipt` is a standalone contract validator).
- `_lifetime_band` guards its `:.1f`/`:.2f` format specs against non-finite ledger fields (`_is_finite_num`).
- `_session_stem` strips control characters (`\x00`/`\n`/…) — closes a silent receipt-drop (null byte) and a newline-in-filename.
- `receipts_enabled` rewritten with explicit falsy-token handling (`_validation._BOOL_FALSE_TOKENS`): unset / empty / `0|false|no|off` → **enabled**; truthy **or any unrecognized value** → **disabled** (privacy fail-safe — an ambiguous opt-out value must never silently start writing).

## Tests

- New `tests/test_dashboard_coercion_corpus.py` — the adversarial corpus (`nan`, `±inf`, `"1e999"`, `10**400`, `""`, whitespace, unicode digits, `bool`, `None`, negative, `list`/`dict`) over every coercion + the `_MAX_RECEIPT_INT` fence-post (`10**15` vs `+1`), the 100%-cap, `_fmt_int` sign, the mixed-`ts` crash, the integer Hamilton, the `allow_nan` boundary, the control-char strip, and the full `receipts_enabled` env matrix. Every bug-capture test is proven **RED at base** v1.8.34.
- Full suite: **1796 passed, 5 skipped** (1711 at base). No change outside the dashboard / metrics / receipts surface (`guard.py` untouched). Verified under a fresh zero-dependency venv.

## Scope

- Observability path only — does not touch the guard / prune / reload core.
- **Intentional behavior change (note for upgraders):** `COZEMPIC_NO_RECEIPTS=0`/`false`/`no`/`off` now **keeps receipts enabled** (it previously disabled them for any non-empty value). Truthy or unrecognized values still disable (privacy fail-safe).
- **Deferred to a follow-up (one bug class per PR):** receipt-file retention/rotation (L3 unbounded growth); `build_receipt` can still serialize a non-portable huge-int JSON number from corrupt input (the read side clamps). `total_strategy_bytes==0` leaves strategies unattributed (designed — byte-share has no denominator).


junaidtitan (approved):
Thorough, well-scoped hardening — thanks. Verified on my side:

- Merges clean into current `main` (1.8.37) and the **merged full suite is green (1859 passed)** — no semantic conflict despite the 1.8.34 base.
- Multi-agent adversarial review found **no real issues**: the integer-Hamilton apportionment still sums exactly to `tokens.reclaimed`, the `_MAX_RECEIPT_INT` fence-posts don't clip any legitimate value, and `validate_receipt`/`serialize_receipt` don't over-reject a valid receipt.
- The `receipts_enabled` change is the right call — the old `not os.environ.get(...)` meant `COZEMPIC_NO_RECEIPTS=false` *disabled* receipts (opposite of intent); the new falsy-token handling + privacy fail-safe fixes that footgun. Documented clearly for upgraders.

Approving. Appreciate the RED-at-base proofs and the honest scope notes.

PR #159, opened 2026-06-29
docs: recommend on-PATH install so the hooks + guard daemon actually start

### LINKED ISSUE #158: Recommended uvx install never lands cozempic on PATH, so the SessionStart guard daemon silently fails (hook still prints "guard active")
## Summary

The README's **recommended** install method is `uvx cozempic`:

```
# uv / uvx — no install needed, run on demand (recommended)
uvx cozempic --help
```

But `uvx` runs on demand in a throwaway environment and **never installs a persistent `cozempic` executable onto `PATH`**. Meanwhile the `SessionStart` hook launches the guard daemon by invoking bare **`cozempic`** (falling back to **`python3 -m cozempic`**) — both of which require cozempic to be on `PATH` or importable by the system `python3`.

So if you follow the **recommended** install, the hook can't find cozempic, the guard daemon **never starts**, and there's no indication anything is wrong — the hook ends with an unconditional `echo 'Cozempic: guard active'` after every launch attempt is swallowed by `2>/dev/null` / `|| true`. The failure only surfaces when you run `/cozempic:doctor`:

```
⚠ cozempic-daemon-running   [WARNING]
    No guard daemon is running. Background protection is inactive for this session.
```

The startup banner and the doctor directly contradict each other, which is what makes this confusing to diagnose.

(Note: the README's other methods behave differently — `pipx install cozempic` is correctly described as "always on PATH" and works; `pip`/`npm -g` work if their bin dir is on `PATH`. The breakage is specific to the **recommended** `uvx` path, which deliberately installs nothing.)

## Environment

- cozempic plugin (hook schema `v5`), bundled CLI/MCP server `1.8.9`
- Recommended install per README: `uvx cozempic`
- `uvx cozempic` resolves to `1.8.37`
- macOS, zsh; `uv`/`uvx` at `/usr/local/bin`
- cozempic **not** on `PATH`; **not** importable by system `python3`

## What the hook actually does

The `SessionStart` command (abridged) is:

```sh
{ cozempic guard --daemon … 2>/dev/null \
  || python3 -m cozempic guard --daemon … 2>/dev/null ; } || true
…
& echo 'Cozempic: guard active'
```

- Branch 1 `cozempic …` → not on `PATH` (uvx installs nothing) → fails silently.
- Branch 2 `python3 -m cozempic …` → module not in system python → fails silently.
- No `uvx` / `uv tool run` fallback exists.
- `echo 'Cozempic: guard active'` runs **unconditionally**, regardless of whether the daemon launched.

The hook's background auto-update step (`pip install --upgrade cozempic`) has the same blind spot — it targets a pip environment that a `uvx`/`uv tool` user doesn't have.

## Steps to reproduce

1. Install via the **recommended** method: `uvx cozempic` (no persistent install).
2. Enable the plugin and start a Claude Code session.
3. Observe `SessionStart` print `Cozempic: guard active`.
4. Run `/cozempic:doctor` → `cozempic-daemon-running [WARNING] No guard daemon is running`.
5. `ps aux | grep "cozempic guard"` → no daemon process.

## Impact

The **recommended** install path produces a plugin that **appears** active on every session but provides **zero** background context protection — and the misleading banner actively hides it. uv is increasingly the default toolchain for Python CLIs, so this likely hits a large share of users who follow the docs as written.

## Suggested fixes

1. **Docs:** stop recommending bare `uvx cozempic` for plugin/daemon use, or clearly flag it as CLI-only. Recommend **`uv tool install cozempic`** instead — it installs a persistent, on-`PATH` executable (uv's tool-bin dir, e.g. `~/.local/bin`) that the hook can actually resolve, while keeping uv as the toolchain. (`pipx install cozempic` is an equally valid on-PATH alternative the README already lists.)
2. **Hook robustness:** add a `uvx cozempic …` / `uv tool run cozempic …` fallback to the daemon-launch (and auto-update) chains so uv-based installs work out of the box.
3. **Honest status:** only print `Cozempic: guard active` when the daemon actually started — gate the `echo` on the launch result, or verify the daemon is alive before claiming it. A silent failure that reports success is worse than a visible error.

## Workaround

```sh
uv tool install cozempic   # persistent `cozempic` on PATH (~/.local/bin)
```

After this, the existing `cozempic guard --daemon` hook branch resolves and the daemon starts on every session.

jordanmiguel:
Fixes #158.

## Problem

The README recommends `uvx cozempic` as the primary install method:

```
# uv / uvx — no install needed, run on demand (recommended)
uvx cozempic --help
```

But `uvx` runs the package in a throwaway environment and **never installs a `cozempic` executable on `PATH`**. The `SessionStart` hook starts the guard daemon by invoking bare `cozempic` (falling back to `python3 -m cozempic`) — neither of which resolves under a `uvx`-only setup. The daemon never starts, yet the hook still prints `Cozempic: guard active` (the `echo` is unconditional), so the failure is invisible until `/cozempic:doctor` reports `No guard daemon is running`. This directly contradicts the README's own promise that the "guard daemon auto-starts on every Claude Code session."

## Change

Docs-only:

- Recommend **`uv tool install cozempic`** (persistent, on `PATH`, still the uv toolchain, and compatible with cozempic's own auto-update) instead of run-on-demand `uvx cozempic`.
- Add a callout right after the install block explaining that the hooks and guard daemon require `cozempic` to be resolvable on `PATH`, why bare `uvx` doesn't satisfy that, and how to verify (`command -v cozempic`).
- In the "As a Claude Code Plugin" section, recommend enabling marketplace **`autoUpdate`** after adding the marketplace. The plugin bundle (hooks/skills/MCP) is versioned separately from the PyPI package and otherwise stays pinned until a manual `/plugin` update — which is how an install can silently drift behind the package.

The other listed methods (`pipx`, `npm -g`, `pip`, Homebrew, Nix) already land an executable on `PATH` and are unaffected.

## Notes / possible follow-ups (out of scope here)

This PR only fixes the docs. Two code-side improvements would make this more robust, happy to open separate issues/PRs if useful:

1. Add a `uvx cozempic …` / `uv tool run cozempic …` fallback to the `SessionStart` daemon-launch chain so a uvx-only setup still works.
2. Only print `Cozempic: guard active` when the daemon actually started — a silent failure that reports success is the part that made this hard to diagnose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,21 @@ Install cozempic (any method above), then inside Claude Code:
 
 This gives you MCP tools, skills (`/cozempic:diagnose`, `/cozempic:treat`, etc.), and auto-wired hooks.
 
+> **Keep the plugin current — enable marketplace auto-update.** The `COZEMPIC_*` controls above govern the PyPI **package** (the `cozempic` CLI + guard daemon). The Claude Code **plugin** (hooks, skills, MCP server) is versioned separately, and unless auto-update is enabled for its marketplace it stays pinned at whatever version you installed until you run `/plugin` by hand. To keep it current automatically, set `autoUpdate` on the marketplace entry in `~/.claude/settings.json`:
+>
+> ```json
+> {
+>   "extraKnownMarketplaces": {
+>     "cozempic": {
+>       "source": { "source": "github", "repo": "Ruya-AI/cozempic" },
+>       "autoUpdate": true
+>     }
+>   }
+> }
+> ```
+>
+> With `autoUpdate: true`, Claude Code refreshes the marketplace **and its installed plugins** on startup.
+
 ## Quick Start
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -32,11 +32,11 @@ Cozempic removes it with **18 composable strategies** across 3 prescription tier
 
 ## Install
 
-Pick your package manager. `uvx`/`pipx`/`npm`/`pip` are the lowest-friction (no Homebrew trust prompt — see the note below):
+Pick your package manager. `uv tool`/`pipx`/`npm`/`pip` are the lowest-friction (no Homebrew trust prompt — see the note below):
 
 ```bash
-# uv / uvx — no install needed, run on demand (recommended)
-uvx cozempic --help
+# uv — persistent install on PATH (recommended); powers the hooks + guard daemon
+uv tool install cozempic
 
 # pipx — isolated user install, always on PATH
 pipx install cozempic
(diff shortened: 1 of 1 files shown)


junaidtitan:
Thanks @jordanmiguel — this is exactly the right diagnosis, and we shipped the same fix (plus the deeper root cause) in **1.8.38**:

- **README**: it now recommends `uv tool install cozempic` as the primary, on-PATH install and explicitly warns that `uvx` is ephemeral and can't keep the guard daemon running (#162, reinforced in #166) — so the text your PR targets no longer exists on `main`.
- **Root cause in code**: #162 also bakes an absolute `<interpreter> -m cozempic` into the hook fallback (`init._bake_cozempic_path`), so the guard resolves even when bare `cozempic` isn't on the hook's PATH — something a docs-only change can't cover.

Closing as superseded, but genuinely appreciated — your report and #158 are what drove the fix. 🙏

PR #167, opened 2026-06-30
fix(team): re-key shared task list by real system id (completed tasks render as pending after compaction)

anasdayeh:
## Problem

After a compaction, the recovered checkpoint's shared task list (`## Active Task List` / "Shared active tasks") renders tasks that were **already completed** as still `[ ]` pending, and sometimes shows phantom blank tasks. Because `cmd_post_compact` writes the checkpoint to stdout (injected as post-compact context) and `inject_team_recovery` injects team state on the prune path, this incorrect list is also fed back to the model — so the agent is told finished work is still outstanding.

## Root cause

`extract_team_state()` keys created and updated tasks in two **different namespaces**:

- `TaskCreate`'s tool **input has no id** — the system assigns it and returns it only in the tool_result (`"Task #N created"`). The extractor therefore keyed created tasks **positionally**: `task_id = _sfield(inp, "taskId", "id") or str(len(seen_tasks))` → `"0", "1", "2", …`.
- `TaskUpdate`'s input **does** carry the real system id (`{"taskId": "1", …}`, 1-based).

So a `TaskUpdate(taskId="1", status="completed")` lands on the **positional** slot `"1"` (the *second* created task), not real task #1. Consequences:

- completions are applied to the wrong task;
- the first-created task (positional key `"0"`) is **never** matched by any update, so it is stuck `pending` forever;
- an update whose id exceeds the positional range falls through to the `else` branch and **fabricates a phantom blank task**.

This is made worse when the task store resets mid-session and re-issues ids from #1 (the positional and real namespaces diverge further).

### Minimal repro

```python
msgs = [
    (0, {"message": {"role": "assistant", "content": [
        {"type": "tool_use", "id": "c1", "name": "TaskCreate", "input": {"subject": "Ship it"}}]}}, 100),
    (1, {"message": {"role": "user", "content": [
        {"type": "tool_result", "tool_use_id": "c1", "content": "Task #1 created successfully: Ship it"}]}}, 80),
    (2, {"message": {"role": "assistant", "content": [
        {"type": "tool_use", "id": "u1", "name": "TaskUpdate", "input": {"taskId": "1", "status": "completed"}}]}}, 80),
]
state = extract_team_state(msgs)
# before: task "Ship it" -> status "pending"  (+ a phantom blank "completed" task)
# after:  task "Ship it" -> status "completed"
```

## Fix

Re-key each `TaskCreate` to its **real** id, parsed from the `"Task #N created"` tool_result:

- park the created task under a temp key tied to its `tool_use_id`;
- when the matching tool_result arrives, parse the real id and re-key the task to it;
- a reused id (task-store reset) overwrites the older generation — latest-generation-wins, matching the live store which only holds the latest.

Fail-open: if a create's tool_result is missing (e.g. a truncated transcript), the task stays visible under its temp key rather than disappearing.

## Tests

`tests/test_task_list_keying.py` (4 cases) — completed-task-not-pending, no-phantom-blank, first-task-not-orphaned, id-reuse-latest-generation-wins. They **fail on the positional-keying base and pass with the fix**. The full existing team suite (`test_team_schema`, `test_guard_team_agent_spawn`, `test_post_compact`, `test_team_recovery_receipt`) stays green — 105 passed total.

### DIFF
--- a/src/cozempic/team.py
+++ b/src/cozempic/team.py
@@ -604,6 +604,13 @@ def extract_team_state(messages: list[Message]) -> TeamState:
     # discard event ordering and mark a still-working teammate "completed").
     last_send_line: dict[str, int] = {}
     seen_tasks: dict[str, TaskInfo] = {}
+    # TaskCreate's tool INPUT carries no id (the system assigns it and returns it
+    # in the tool_result: "Task #N created"). Map each TaskCreate tool_use_id to
+    # the temp key its task is parked under, so the result handler can re-key it
+    # to the REAL id. Without this, creates were keyed positionally (str(len(...)))
+    # while TaskUpdate keys by the real id → completed tasks render as pending and
+    # phantom blank tasks appear in the recovered checkpoint.
+    taskcreate_uid_to_key: dict[str, str] = {}
 
     # Track tool_use_id -> tool_name for matching results to calls
     tool_use_id_to_name: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


junaidtitan:
Reviewed with our full find→adversarial-verify fleet + independent empirical checks against real transcripts. **Root cause and fix both confirmed** — the old positional key (`str(len)`, 0-based) never matched `TaskUpdate.taskId` (1-based system id), so *every* normal completion forked a phantom blank; this makes the well-formed case correct for the first time. Real result format `"Task #N created successfully"` verified across #1–#28+ in live logs.

Pushed three torn-transcript hardening nits on top (all with regression tests, full suite green at 1889):
1. Consume the uid→tempkey mapping **only on a successful regex match** — a reworded/torn first result no longer strands the task; a later duplicate good result recovers the id.
2. On re-key collision, distinguish a real prior-generation task (has a subject → store-reset id reuse, latest wins) from a subject-less out-of-order `TaskUpdate` fragment (keep its authoritative status/owner, backfill subject) — so a completed task is never re-rendered as pending.
3. Normalize any lingering `__pending_create_<uid>` sentinel to a clean positional id at finalize so it never leaks to JSON consumers.

Thanks @anasdayeh — clean, well-targeted fix. Merging.

PR #177, opened 2026-07-15
fix(init): refresh stale global hooks after upgrades

eggrollofchaos:
## Summary

- Refresh existing stale global Cozempic hooks even when the global-init marker is present.
- Preserve the marker-without-hooks path as a prior user opt-out.
- Cover migration of a legacy schema-v10 global hook without prompting.

## Why

A package upgrade can ship a safer SessionStart hook while an earlier globally installed hook remains in `~/.claude/settings.json`. The marker previously suppressed the refresh path, leaving the old hook active.

## Tests

- `uv run --frozen --with pytest pytest tests/test_global_init.py tests/test_auto_init_global_skip.py -q`
- `uv run --frozen --with pytest pytest tests/test_guard_race_2026_05_18.py::TestR1_DaemonProcessRace::test_two_processes_one_winner -q`
- `uv run --frozen --with pytest pytest -x -q` reached 791 passed and 5 skipped before the existing nondeterministic guard-race test produced two winners in one of 50 iterations; its isolated rerun passed.

## Review

Required review tiers: tier-1,tier-2,tier-3

Approval authority: requester

### DIFF
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1204,7 +1204,7 @@ def cmd_doctor(args):
 def cmd_uninstall(args):
     """Reverse `cozempic init` — remove hooks, the slash command, and markers.
 
-    Default scope is GLOBAL (~/.claude). Keeps the user's data (savings ledger,
+    Default scope is Claude's configured global directory. Keeps the user's data (savings ledger,
     receipts) unless --purge. Sets the auto-init opt-out so init won't re-wire.
     """
     from .init import preview_uninstall, run_uninstall
@@ -1223,6 +1223,8 @@ def cmd_uninstall(args):
         print(f"    Slash command (~/.claude/commands/cozempic.md): "
               f"{'remove' if prev['slash_command'] else '(not present / not ours)'}")
         print(f"    Remind counter: {'remove' if prev['remind_counter'] else '(none)'}")
+        for error in prev.get("errors", []):
+            print(f"    Hooks: ERROR — {error}")
         if purge:
             print(f"    Purge data: {prev['purge_data'] or '(none)'}")
         print()
@@ -1240,11 +1242,14 @@ def cmd_uninstall(args):
 
     result = run_uninstall(scope, purge)
     n_hooks = sum(len(h.get("removed", [])) for h in result["hooks"])
-    print(f"  Removed {n_hooks} hook(s) across {len(result['hooks'])} settings file(s).")
+    successful_settings = [h for h in result["hooks"] if not h.get("error")]
+    print(f"  Removed {n_hooks} hook(s) across {len(successful_settings)} settings file(s).")
     for h in result["hooks"]:
         if h.get("removed"):
             print(f"    - {h['settings_path']}: {', '.join(h['removed'])}"
                   + (f"  (backup: {h['backup_path']})" if h.get("backup_path") else ""))
+        elif h.get("error"):
+            print(f"    - ERROR: {h['error']}")
     sc = result.get("slash_command")
     if sc and sc.get("removed"):
         print(f"  Removed slash command: {sc['path']}"
@@ -1264,10 +1269,13 @@ def cmd_init(args):
         # Deprecated alias → `cozempic uninstall --global`.
         print("  Note: `init --uninstall-global` is deprecated; use `cozempic uninstall`.", file=sys.stderr)
         from .init import uninstall_hooks
-        result = uninstall_hooks(str(Path.home()))
+        global_settings = _global_claude_dir() / "settings.json"
+        result = uninstall_hooks(str(Path.home()), settings_path=global_settings)
         print("\n  COZEMPIC INIT — UNINSTALL GLOBAL")
         print("  ═══════════════════════════════════════════════════════════════════")
-        if result.get("removed"):
+        if result.get("error"):
+            print(f"  Hooks: ERROR — {result['error']}")
+        elif result.get("removed"):
             print(f"  Removed {len(result['removed'])} hook(s) from {result['settings_path']}")
             for h in result["removed"]:
                 print(f"    - {h}")
@@ -1275,17 +1283,19 @@ def cmd_init(args):
                 print(f"  Backup: {result['backup_path']}")
         else:
             print("  No cozempic hooks found in ~/.claude/settings.json — nothing to remove.")
-        # Mark as opted-out so global auto-init doesn't re-fire
-        try:
-            _GLOBAL_INIT_MARKER.touch()
-        except OSError:
-            pass
+        if not result.get("error"):
+            # Mark as opted-out so global auto-init doesn't re-fire.
+            try:
+                _GLOBAL_INIT_MARKER.touch()
+            except OSError:
+                pass
         print()
         return
 
     if getattr(args, "global_install", False):
         project_dir = str(Path.home())
-        scope_label = "GLOBAL (~/.claude/)"
+        global_settings = _global_claude_dir() / "settings.json"
+        scope_label = f"GLOBAL ({global_settings.parent}/)"
     else:
         project_dir = args.cwd or os.getcwd()
         scope_label = f"Project: {project_dir}"
@@ -1295,13 +1305,11 @@ def cmd_init(args):
     print(f"  {scope_label}")
     print()
 
-    result = run_init(project_dir, skip_slash=args.no_slash_command)
-    if getattr(args, "global_install", False):
-        # Mark as initialized so global auto-init doesn't re-fire
-        try:
-            _GLOBAL_INIT_MARKER.touch()
-        except OSError:
-            pass
+    result = run_init(
+        project_dir,
+        skip_slash=args.no_slash_command,
+        settings_path=global_settings if getattr(args, "global_install", False) else None,
+    )
 
     # Report hooks
     hooks = result["hooks"]
@@ -1322,6 +1330,13 @@ def cmd_init(args):
     if hooks["skipped"]:
         for h in hooks["skipped"]:
             print(f"    ~ {h} (current, skipped)")
+
+    if getattr(args, "global_install", False) and not hooks.get("error"):
+        # A failed install must remain retryable instead of looking like a decline.
+        try:
+            _GLOBAL_INIT_MARKER.touch()
+        except OSError:
+            pass
 
     # #158: if cozempic is running from a throwaway env (uvx / uv run), the path
     # baked into the hooks won't exist next session, so the guard daemon can't
@@ -2057,16 +2072,16 @@ def _prompt_with_timeout(msg: str, timeout: int = 30, default: str = "n") -> str
 
 
 def _maybe_global_init(argv: list[str]) -> None:
-    """Wire cozempic into ~/.claude/settings.json on first cozempic invocation
+    """Wire cozempic into Claude's configured global settings on first invocation
     on this machine — guarantees protection for every Claude Code session in
     every project, including projects the user has never run cozempic in.
 
     Bail-outs (in order):
       1. COZEMPIC_NO_GLOBAL_INIT=1 in env (also set by --no-global-init)
       2. Subcommand is in _AUTO_INIT_SKIP_CMDS
-      3. ~/.claude/ doesn't exist (Claude Code not installed yet)
+      3. Claude's configured global directory doesn't exist yet
       4. Marker exists and no Cozempic hooks are present (the user declined)
-      5. Cozempic hooks are already current in ~/.claude/settings.json (e.g. plugin
+      5. Cozempic hooks are already current in Claude's global settings (e.g. plugin
          marketplace install)
 
     Otherwise: writes hooks (skip slash command — project-level only) and
@@ -2091,7 +2106,7 @@ def _maybe_global_init(argv: list[str]) -> None:
     if cmd is None and not is_version_check:
         return
 
-    home_claude = Path.home() / ".claude"
+    home_claude = _global_claude_dir()
     if not home_claude.exists():
         return  # Claude Code not yet installed — defer until it is
 
@@ -2100,7 +2115,7 @@ def _maybe_global_init(argv: list[str]) -> None:
     if hook_state.error:
         print(
             f"  Cozempic: global init FAILED — {hook_state.error}. "
-            "Fix ~/.claude/settings.json, then run `cozempic init --global`.",
+            f"Fix {home_claude / 'settings.json'}, then run `cozempic init --global`.",
             file=sys.stderr,
         )
         return
@@ -2137,7 +2152,7 @@ def _maybe_global_init(argv: list[str]) -> None:
                 file=sys.stderr,
             )
             print(
-                "  Wires hooks into ~/.claude/settings.json. Reverse any time with "
+                "  Wires hooks into Claude's global settings. Reverse any time with "
                 "`cozempic init --uninstall-global`.",
                 file=sys.stderr,
             )
@@ -2159,7 +2174,11 @@ def _maybe_global_init(argv: list[str]) -> None:
             return
 
     try:
-        result = run_init(str(Path.home()), skip_slash=True)
+        result = run_init(
+            str(Path.home()),
+            skip_slash=True,
+            settings_path=home_claude / "settings.json",
+        )
     except Exception as exc:
         print(
             f"  Cozempic: global init failed ({exc}). Run `cozempic init --global` manually "
@@ -2236,6 +2255,11 @@ def _managed_cozempic_hook_state(claude_dir: Path):
     return cozempic_hook_schema_state(claude_dir / "settings.json")
 
 
+def _global_claude_dir() -> Path:
+    from .session import get_claude_dir
+    return get_claude_dir()
+
+
 def _project_cozempic_hook_state(claude_dir: Path):
     """Return combined state for the project's effective Claude settings files."""
     from .init import cozempic_hook_schema_state_for_paths
@@ -2272,13 +2296,13 @@ def _maybe_auto_init(argv: list[str]) -> None:
     # Skip local init when global hooks are already current (avoids double-firing).
     # Guard: if cwd IS the home dir, home_claude == claude_dir -- fall through
     # to the normal local check instead.
-    home_claude = Path.home() / ".claude"
+    home_claude = _global_claude_dir()
     if home_claude != claude_dir:
         global_state = _managed_cozempic_hook_state(home_claude)
         if global_state.error:
             print(
                 f"  Cozempic: auto-init skipped ({global_state.error}). "
-                "Fix ~/.claude/settings.json, then run `cozempic init --global`.",
+                f"Fix {home_claude / 'settings.json'}, then run `cozempic init --global`.",
                 file=sys.stderr,
             )
             return
@@ -2293,7 +2317,15 @@ def _maybe_auto_init(argv: list[str]) -> None:
                 )
             return
 
-    if _project_is_cozempic_current(claude_dir):
+    project_state = _project_cozempic_hook_state(claude_dir)
+    if project_state.error:
+        print(
+            f"  Cozempic: auto-init skipped ({project_state.error}). "
+            "Fix the project settings, then run `cozempic init`.",
+            file=sys.stderr,
+        )
+        return
+    if project_state.found and project_state.current:
         return  # already initialized
 
     try:
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -2063,10 +2063,10 @@ def _maybe_global_init(argv: list[str]) -> None:
 
     Bail-outs (in order):
       1. COZEMPIC_NO_GLOBAL_INIT=1 in env (also set by --no-global-init)
-      2. Marker file exists (already done once)
-      3. Subcommand is in _AUTO_INIT_SKIP_CMDS
-      4. ~/.claude/ doesn't exist (Claude Code not installed yet)
-      5. Cozempic hooks are already in ~/.claude/settings.json (e.g. plugin
+      2. Subcommand is in _AUTO_INIT_SKIP_CMDS
+      3. ~/.claude/ doesn't exist (Claude Code not installed yet)
+      4. Marker exists and no Cozempic hooks are present (the user declined)
+      5. Cozempic hooks are already current in ~/.claude/settings.json (e.g. plugin
          marketplace install)
 
     Otherwise: writes hooks (skip slash command — project-level only) and
@@ -2074,8 +2074,6 @@ def _maybe_global_init(argv: list[str]) -> None:
     we never ask again on this machine.
     """
     if os.environ.get("COZEMPIC_NO_GLOBAL_INIT"):
-        return
-    if _GLOBAL_INIT_MARKER.exists():
         return
 
     # --help / -h are pure-info, never trigger init
@@ -2097,6 +2095,7 @@ def _maybe_global_init(argv: list[str]) -> None:
     if not home_claude.exists():
         return  # Claude Code not yet installed — defer until it is
 
+    marker_exists = _GLOBAL_INIT_MARKER.exists()
     if _project_is_cozempic_current(home_claude):
         # Already wired (probably via plugin marketplace install). Mark as done.
         try:
@@ -2105,10 +2104,17 @@ def _maybe_global_init(argv: list[str]) -> None:
             pass
         return
 
+    if marker_exists and not _project_has_cozempic_hooks(home_claude):
+        # A marker without hooks records a prior explicit decline. Do not turn a
+        # future package upgrade into an unsolicited first install.
+        return
+
     # Ask the user interactively when both stdin and stderr are TTYs (real terminal).
     # Fall back to silent auto-install for non-interactive contexts (CI, pipelines,
     # Claude Code subprocess invocations) so we never hang waiting for input.
-    interactive = sys.stdin.isatty() and sys.stderr.isatty()
+    # Existing but stale hooks already establish consent; refresh them without
+    # prompting so an automatic package update also advances the hook schema.
+    interactive = not marker_exists and sys.stdin.isatty() and sys.stderr.isatty()
 
     if interactive:
         try:
@@ -2215,10 +2221,23 @@ def _project_is_cozempic_current(claude_dir: Path) -> bool:
     query. If one file is current and another has stale cozempic hooks, we
     return False so wire_hooks is called and refreshes the stale one.
     """
+    found, current = _project_cozempic_hook_state(claude_dir)
+    return found and current
+
+
+def _project_has_cozempic_hooks(claude_dir: Path) -> bool:
+    """Return whether either global settings file contains Cozempic hooks."""
+    found, _ = _project_cozempic_hook_state(claude_dir)
+    return found
+
+
+def _project_cozempic_hook_state(claude_dir: Path) -> tuple[bool, bool]:
+    """Return ``(has_cozempic_hooks, all_cozempic_hooks_are_current)``."""
     import json as _json
     from .init import _is_cozempic_command, HOOK_SCHEMA_MARKER
 
     any_cozempic_found = False
+    any_stale_cozempic_found = False
     for name in ("settings.json", "settings.local.json"):
         p = claude_dir / name
         if not p.exists():
@@ -2248,9 +2267,9 @@ def _project_is_cozempic_current(claude_dir: Path) -> bool:
                     # Any non-current cozempic hook (missing or stale marker)
                     # means a refresh is due.
                     if HOOK_SCHEMA_MARKER not in cmd:
-                        return False
+                        any_stale_cozempic_found = True
 
-    return any_cozempic_found
+    return any_cozempic_found, not any_stale_cozempic_found
 
 
 def _maybe_auto_init(argv: list[str]) -> None:
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1259,10 +1259,18 @@ def cmd_uninstall(args):
               + (f"  (backup: {sc['backup_path']})" if sc.get("backup_path") else ""))
     elif sc and sc.get("skipped_foreign"):
         print(f"  Left {sc['path']} in place (not a cozempic command).")
+    elif sc and sc.get("error"):
+        print(f"  Slash command: ERROR — {sc['error']}")
     if result.get("purged"):
         print(f"  Purged data: {', '.join(result['purged'])}")
+    reported_hook_errors = {h.get("error") for h in result["hooks"] if h.get("error")}
+    for error in result.get("errors", []):
+        if error not in reported_hook_errors:
+            print(f"  ERROR: {error}")
     if result.get("opt_out_set"):
         print("  Auto-init disabled. Re-run `cozempic init` to reinstall.")
+    if result.get("project_opt_out_set"):
+        print("  Project auto-init disabled. Re-run `cozempic init` to reinstall.")
     print()
 
 
@@ -1288,10 +1296,7 @@ def cmd_init(args):
             print("  No cozempic hooks found in ~/.claude/settings.json — nothing to remove.")
         if not result.get("error"):
             # Mark as opted-out so global auto-init doesn't re-fire.
-            try:
-                _global_init_marker().touch()
-            except OSError:
-                pass
+            _persist_global_init_marker()
         print()
         return
 
@@ -1335,6 +1340,8 @@ def cmd_init(args):
     if hooks["skipped"]:
         for h in hooks["skipped"]:
             print(f"    ~ {h} (current, skipped)")
+    for warning in hooks.get("warnings", []):
+        print(f"  Hooks: WARNING — {warning}")
 
     local_cleanup = result.get("local_cleanup")
     if local_cleanup:
@@ -1348,10 +1355,7 @@ def cmd_init(args):
 
     if getattr(args, "global_install", False) and not hooks.get("error"):
         # A failed install must remain retryable instead of looking like a decline.
-        try:
-            _global_init_marker().touch()
-        except OSError:
-            pass
+        _persist_global_init_marker()
 
     # #158: if cozempic is running from a throwaway env (uvx / uv run), the path
     # baked into the hooks won't exist next session, so the guard daemon can't
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1319,12 +1319,14 @@ def cmd_init(args):
     updated = hooks.get("updated") or []
     if hooks.get("error"):
         print(f"  Hooks: ERROR — {hooks['error']}")
-    elif hooks["added"] or updated:
+    elif hooks["added"] or updated or hooks.get("repaired"):
         print(f"  Hooks wired in {hooks['settings_path']}:")
         for h in hooks["added"]:
             print(f"    + {h} (added)")
         for h in updated:
             print(f"    → {h} (refreshed from stale schema)")
+        if hooks.get("repaired"):
+            print("    → malformed hook entries removed")
         if hooks["backup_path"]:
             print(f"  Backup: {hooks['backup_path']}")
     else:
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1154,6 +1154,7 @@ def cmd_doctor(args):
         "warning": "⚠",
         "issue": "✗",
         "fixed": "→",
+        "error": "!",
     }
     STATUS_COLORS = {
         "ok": "",
@@ -1171,6 +1172,7 @@ def cmd_doctor(args):
     issues = 0
     warnings = 0
     fixed = 0
+    errors = 0
 
     for r in results:
         icon = STATUS_ICONS.get(r.status, "?")
@@ -1188,12 +1190,14 @@ def cmd_doctor(args):
             warnings += 1
         elif r.status == "fixed":
             fixed += 1
+        elif r.status == "error":
+            errors += 1
 
     # Summary
     if fixed:
         print(f"  Summary: {fixed} issue(s) fixed")
-    elif issues or warnings:
-        print(f"  Summary: {issues} issue(s), {warnings} warning(s)")
+    elif issues or warnings or errors:
+        print(f"  Summary: {issues} issue(s), {warnings} warning(s), {errors} check(s) failed")
         if not args.fix:
             print("  Run 'cozempic doctor --fix' to auto-fix where possible.")
     else:
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1257,7 +1257,7 @@ def cmd_uninstall(args):
             purge = False
         elif _prompt_with_timeout("  Continue? [y/N] ", timeout=30, default="n") != "y":
             print("  Aborted.\n")
-            return
+            raise SystemExit(1)
 
     result = run_uninstall(scope, purge)
     n_hooks = sum(len(h.get("removed", [])) for h in result["hooks"])
--- a/src/cozempic/cli.py
+++ b/src/cozempic/cli.py
@@ -1222,7 +1222,7 @@ def cmd_uninstall(args):
 
     if purge and scope == "project":
         print("  ERROR: --purge only applies to global data; omit --project or use --all.\n")
-        return
+        raise SystemExit(2)
 
     if getattr(args, "dry_run", False):
         prev = preview_uninstall(scope, purge)
@@ -1243,6 +1243,8 @@ def cmd_uninstall(args):
         if purge:
             print(f"    Purge data: {', '.join(prev['purge_data']) or '(none)'}")
         print()
+        if prev["hook_errors"] or prev.get("slash_error"):
+            raise SystemExit(1)
         return
 
     if purge:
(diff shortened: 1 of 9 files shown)


eggrollofchaos (commented):
Review result at exact head `0dcf2ba06af7a77c74d51ce977ecd35064881b38`.

## Critical

1. `src/cozempic/cli.py:2098-2117, 2234-2272` — the new admission decision scans both `settings.json` and `settings.local.json`, but the selected mutator (`run_init` → `wire_hooks`) only refreshes `settings.json`. This makes a stale Cozempic hook in `settings.local.json` permanently unrefreshable. More seriously, `src/cozempic/init.py:636-666` promises that the marker left by global uninstall prevents future rewiring, but uninstall also removes hooks only from `settings.json`. If a stale hook remains in `settings.local.json`, the next ordinary command treats it as installation consent and silently recreates all global hooks in `settings.json`, reversing the explicit uninstall. I reproduced both paths: a local-only stale hook remained stale after repeated `_maybe_global_init()` calls, and an uninstall followed by `cozempic list` recreated global hooks while state stayed `(found=True, current=False)`. Fix the contract categorically: every file used to decide freshness/consent must be covered by refresh and uninstall, or exclude `settings.local.json` from that decision and preserve an explicit uninstall/decline state that residual hooks cannot override. Add regression tests for stale-only `settings.local.json`, mixed current/stale files, duplicate same-matcher entries, and uninstall followed by a normal command.

## High

None.

## Medium

1. `src/cozempic/cli.py:2234-2272` — the new hook-state traversal has cognitive complexity 40 (the PR's Sonar annotation; threshold 15). This is a real maintenance risk on the exact logic that now distinguishes current, stale, decline, and uninstall states. Move the settings/hook iterator into a small shared helper in `init.py` and have detection, refresh, and uninstall consume the same normalized records. The Critical fix above should make this reduction natural rather than cosmetic.

## Low

None.

## Nit

None.

Verification: `env -u COZEMPIC_NO_AUTO_INIT uv run --frozen --with pytest pytest tests/test_global_init.py tests/test_auto_init_global_skip.py -q` → 31 passed, 2 subtests passed; the isolated guard-race test passed; `git diff --check` passed. The unmodified ambient command initially failed four auto-init tests only because this reviewer environment exports `COZEMPIC_NO_AUTO_INIT`; rerunning with the variable removed matched the packet evidence. Two adversarial tempfile probes reproduced the Critical state-machine gap.

Counts: Critical 1 | High 0 | Medium 1 | Low 0 | Nit 0

Verdict: Critical/High findings remain.

Merge blocker: the Critical uninstall/refresh contract violation remains.

eggrollofchaos (commented):
Review verdict: Critical/High findings remain.

## High

1. `src/cozempic/cli.py:2234-2238, 2280` — the new settings.json-only helper is shared by global init and project auto-init. That fixes the global uninstall contract, but it regresses project detection: a project whose effective Cozempic hook exists only in `.claude/settings.local.json` is now reported as uninitialized, so `_maybe_auto_init()` calls `run_init()` and creates a second managed hook set in `.claude/settings.json`. The base implementation scanned both files, and an adversarial probe at this head confirmed `run_init()` is called for a current local-only hook. This can double-fire the guard/checkpoint hooks. Split the predicates: global init/uninstall should inspect only the managed `settings.json`, while project auto-init should continue recognizing current effective hooks in either project settings file. Add a regression test for current hooks in project `settings.local.json` with no project `settings.json`.

2. `src/cozempic/cli.py:2107-2110` and `src/cozempic/init.py:139-144` — a marker plus malformed/unreadable `settings.json` is silently classified as an explicit decline. `cozempic_hook_schema_state()` collapses parse failure to `(False, True)`, `_project_has_cozempic_hooks()` returns false, and `_maybe_global_init()` returns before `wire_hooks()` can surface its parse error. I reproduced this with a marker and `{broken json`: the command emitted no failure and never called `run_init()`, leaving global protection silently absent. Preserve an error/unknown state instead of equating it with “no hooks”; fail safely without rewriting the file, but surface the parse failure and retry guidance. Add a marker-plus-malformed-settings regression test.

## Medium

1. `src/cozempic/init.py:139-168` — the extracted schema-state scanner still has cognitive complexity 27 (the current Sonar annotation; threshold 15). This is the same load-bearing state parser that now controls current/stale/declined decisions. Extract a normalized iterator over hook commands (or reuse one shared with uninstall/wiring) so malformed shapes and managed-command classification are handled once.

## Low

1. `src/cozempic/cli.py:2115-2117` — the comment says existing stale hooks establish consent, but `interactive` only checks marker presence. A pre-marker installation with stale hooks still receives the first-install prompt; answering “no” touches the marker, then the next invocation refreshes those hooks anyway because marker + hooks is treated as prior consent. Include `_project_has_cozempic_hooks(home_claude)` in the prompt gate so the behavior matches the documented state machine.

## Nit

None.

Verification: `env -u COZEMPIC_NO_AUTO_INIT -u COZEMPIC_NO_GLOBAL_INIT uv run --frozen --with pytest pytest tests/test_global_init.py tests/test_uninstall.py tests/test_auto_init_global_skip.py -q --rootdir <worktree>` → 44 passed, 2 subtests passed; `git diff --check` passed. Two isolated probes reproduced the local-only duplicate-init path and the marker-plus-malformed-settings error masking.

PR context checked: head `5fc98f9230decda3652e2153cf2ce6bfd76ee5d0`; top-level comments 1; inline comments 0; review threads 0 (0 active); prior review records 1; Sonar check passed with 1 current annotation, accepted above as Medium; latest review/comment at 2026-07-15T06:54:02Z.

Counts: Critical 0 | High 2 | Medium 1 | Low 1 | Nit 0

Merge blocker: the two High state-classification regressions remain.

eggrollofchaos (commented):
Self-review verdict: Critical/High findings remain.

## High

1. `src/cozempic/cli.py:2280-2295` — the global-current skip still calls the project-wide predicate for `~/.claude`, so unmanaged `~/.claude/settings.local.json` state can override a current managed `settings.json`. At this exact head, a current global `settings.json` plus a stale global `settings.local.json` makes the combined predicate false; `_maybe_auto_init()` then writes a complete hook set into the current project even though the managed global hooks are already active. The exact-head probe created local PostCompact/PostToolUse/PreCompact/SessionStart/Stop hooks, producing redundant global + project firing. The malformed-local sibling path also collapses to false. Use `_managed_cozempic_hook_state(home_claude)` (including its error handling) for the global skip, and reserve `_project_cozempic_hook_state()` for the actual project. Add mixed current-global/stale-global-local and malformed-global-local regressions.

2. `src/cozempic/cli.py:2161-2175` and `tests/test_global_init.py:621-636` — an operational `run_init()` failure is still converted into the same marker state as an explicit decline. An exact-head `PermissionError` probe showed the first attempt touched the marker and the second invocation silently skipped `run_init()`, so a transient read-only/permission/disk failure permanently disables automatic protection until the user discovers and deletes the marker. That directly violates the requested “failure without becoming an opt-out” state. Keep explicit decline/uninstall distinct from retry throttling (for example, a separate failure/backoff record), and replace the test that currently requires the marker on failure with a retry/non-opt-out regression.

## Medium

None.

## Low

None.

## Nit

None.

Verification: focused exact-head suite passed (47 tests, 2 subtests); `git diff --check` passed. Adversarial tempfile probes reproduced both High paths, including creation of seven local hook entries and failure-marker suppression on the second invocation. PR context checked: head `d7415913b5c174b40a56739d2378c9f3daab4b4c`; 1 top-level bot comment; 0 inline comments; 0 review threads; Sonar check green with 0 new issues.

Counts: Critical 0 | High 2 | Medium 0 | Low 0 | Nit 0

Verdict: Critical/High findings remain.

Approval workflow: not-authority. No approval or merge was attempted.

eggrollofchaos (commented):
## Findings

- Critical — `src/cozempic/init.py:827-846,898-924`: a global slash-command cleanup failure is not included in `global_cleanup_failed`. `run_uninstall("global", purge=True)` reports that failure but still deletes `~/.cozempic`; the preview also says the data is purgeable. A controlled probe with a slash backup failure produced `purge_skipped=False` and removed the data directory. Treat slash read/backup/removal failures as global cleanup failures in both paths, and cover each with purge regression tests.

- High — `tests/test_auto_init_global_skip.py:13-18,111-331`: the changed global-init tests do not isolate their configured Claude directory from the session-backed global directory path. The exact-head targeted suite fails seven assertions. Fix the fixture through the production directory seam, then preserve the new behavior assertions.

## Verification

`uv run --frozen pytest -q tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` → 7 failed, 170 passed, 2 subtests passed.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 1 | High 1 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain.
- Final-head clearance: blocked: Critical/High findings remain.
- Approval workflow: not-authority

eggrollofchaos (commented):
## Review summary

No Critical or High findings.

Verification: `env -u COZEMPIC_NO_AUTO_INIT -u COZEMPIC_NO_AUTO_UPDATE uv run --frozen pytest -q --rootdir <reviewed checkout> tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` — 177 passed, 2 subtests passed. `git diff --check` passed.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 0 | High 0 | Medium 0 | Low 0 | Nit 0
- Verdict: LGTM — approve
- Final-head clearance: tier-3 clean at the reviewed target
- Approval workflow: not-authority

eggrollofchaos (commented):
## Findings

### High

1. `src/cozempic/init.py:898-924` — a slash-command cleanup failure does not set `global_cleanup_failed`, so a confirmed `cozempic uninstall --purge` still deletes `~/.cozempic` and the savings ledger after `uninstall_slash_command()` fails. The dry-run path has the same gate: it reports a slash error but still advertises the purge. Reproduction: make an owned global slash command, force its backup copy to raise `OSError`, then call `run_uninstall("global", purge=True)`; the result contains the backup error with `purge_skipped == False`, and the data directory is gone. Treat every global cleanup failure, including slash-command failures, as purge-blocking in both live and preview paths, and add this regression.

2. `tests/test_auto_init_global_skip.py:13-18` — the fixture patches `cozempic.session.get_claude_dir` with a lambda that uses the test module's `Path.home()`, while each test changes only `cli.Path.home()` to its temporary home. The new configured-profile lookup therefore escapes the fixture. Seven tests in this changed file now fail at this head, including the absent/stale/global-malformed and redundancy paths. Patch the configured-directory seam to the same temporary profile used by the test, then run the full changed test set.

## Verification

- `uv run --frozen pytest -q tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` → 7 failed, 170 passed, 2 subtests passed; the seven failures are in `test_auto_init_global_skip.py`.
- Exact-head temporary reproduction: forced slash-command backup failure returned an error but removed the purge data directory.
- `git diff --check fd41e95b85f283286eaca9419c19825abb6d053b 7924c74ce21ed88d731f8f95816176075d6133ba` → passed.
- CI status: SonarCloud Code Analysis passed; no test check is reported.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 0 | High 2 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain
- Final-head clearance: blocked — High findings remain
- Approval workflow: not-authority
- Merge blocker: the two High findings remain

eggrollofchaos (commented):
## Findings

### Critical

1. `src/cozempic/init.py:827-846,898-924` — a global slash-command cleanup failure is excluded from `global_cleanup_failed`. `run_uninstall("global", purge=True)` reports the slash error but still deletes `~/.cozempic`; the dry run also advertises that data as purgeable. Treat slash read, backup, and removal failures as global-cleanup failures in both paths, and add purge regression coverage.

### High

1. `tests/test_auto_init_global_skip.py:13-18,111-331` — the changed global-init tests do not isolate their configured Claude directory from the new session-backed global-directory path. The exact-head targeted suite fails seven assertions. Inject the intended temporary global directory through the production seam, then retain the assertions for the new behavior.

## Verification

- `uv run --frozen pytest -q tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` → 7 failed, 170 passed, 2 subtests passed.
- Controlled purge probe with a slash-backup failure → error reported, `purge_skipped=False`, data directory removed.
- `git diff --check fd41e95b85f283286eaca9419c19825abb6d053b..7924c74ce21ed88d731f8f95816176075d6133ba` → passed.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 1 | High 1 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain.
- Final-head clearance: blocked: Critical/High findings remain.
- Approval workflow: not-authority

eggrollofchaos (commented):
## Review correction

This supersedes my prior clean review at this head.

## Critical

1. `src/cozempic/init.py:827-846, 898-924` — `--purge` only fails closed for settings-file errors. If removal of the managed global slash command fails (for example, its backup cannot be written), `run_uninstall("global", purge=True)` records the error but still deletes `~/.cozempic`; the dry-run reports that same data as purgeable. A controlled probe with a failing slash backup returned `purge_skipped=False` and removed the data directory. Treat global slash read/backup/removal failures as global cleanup failures in both `preview_uninstall()` and `run_uninstall()`, then add regression coverage for the failure class.

Verification: focused suite with the packet's documented environment controls — 177 passed, 2 subtests passed. A direct purge probe reproduced the data deletion after a slash-backup failure. `git diff --check` passed.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: not cleared; reviewed base fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: not cleared; reviewed target 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 1 | High 0 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain
- Final-head clearance: blocked: Critical finding remains
- Approval workflow: not-authority

eggrollofchaos (commented):
## Findings

### Critical

1. `src/cozempic/init.py:827-846,898-924` — a global slash-command cleanup failure is excluded from `global_cleanup_failed`. Consequently, `run_uninstall("global", purge=True)` reports the slash error but still deletes `~/.cozempic`; `preview_uninstall(..., purge=True)` also advertises the data as purgeable. A controlled exact-head probe with `uninstall_slash_command()` returning `{"error": "backup failed"}` produced `purge_skipped=False` and removed the data directory. Treat slash read/backup/removal failures as global cleanup failures in both preview and live paths, then add purge regression coverage for each failure form.

### High

1. `tests/test_auto_init_global_skip.py:13-18,111-331` — the changed global-init tests do not isolate their configured Claude directory from the new session-backed global directory path. The exact-head targeted suite fails seven assertions (including absent/stale global hooks, managed-malformed settings, cleanup-error reporting, and global/project redundancy). Fix the fixture to inject the intended temporary global directory through the production seam, then keep the assertions that exercise the new behavior.

### Verification

- `uv run --frozen pytest -q tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` → 7 failed, 170 passed, 2 subtests passed.
- Controlled `run_uninstall("global", purge=True)` probe with slash backup failure → error reported, `purge_skipped=False`, data directory removed.
- `git diff --check fd41e95b85f283286eaca9419c19825abb6d053b..7924c74ce21ed88d731f8f95816176075d6133ba` → passed.
- SonarCloud check passing; no current review-thread or inline-comment findings.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 1 | High 1 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain.
- Approval workflow: not-authority

eggrollofchaos (commented):
Correction to [the earlier review](https://github.com/Ruya-AI/cozempic/pull/177#pullrequestreview-4784861433): the first finding is **Critical**, not High.

## Critical

1. `src/cozempic/init.py:827-846,898-924` — a slash-command read, backup, or removal failure is not included in `global_cleanup_failed`. A confirmed global purge can therefore delete `~/.cozempic` and the savings ledger after part of global cleanup failed; dry-run makes the same unsafe purge decision. Reproduction at this head forced the slash backup to raise `OSError`: the command reported the error with `purge_skipped == False`, then removed the data directory. Block purge on every global cleanup failure in both paths and add coverage for all slash failure forms.

## High

1. `tests/test_auto_init_global_skip.py:13-18,111-331` — the configured-directory fixture no longer follows each test's temporary home, so seven cases in this changed test file fail at the target head. Patch the configured-directory seam to use the temporary profile and rerun the changed test set.

## Verification

- `uv run --frozen pytest -q tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py` → 7 failed, 170 passed, 2 subtests passed.
- Exact-head slash-backup failure probe reproduced purge despite the reported cleanup error.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 7924c74ce21ed88d731f8f95816176075d6133ba
- Counts: Critical 1 | High 1 | Medium 0 | Low 0 | Nit 0
- Verdict: Critical/High findings remain
- Final-head clearance: blocked — Critical/High findings remain
- Approval workflow: not-authority
- Merge blocker: the Critical and High findings remain

eggrollofchaos (commented):
Reviewed at `7924c74c`. I probed each of the four stated guarantees directly rather than reading for them, building every fixture with the real installer so nothing was tested against a hand-written file the code would correctly treat as foreign.

## Critical — `--purge` deletes user data after a failed slash-command cleanup

`run_uninstall` records a slash-command failure in `result["errors"]` but never feeds it into `global_cleanup_failed`, so the purge guard does not see it:

```python
if scope in ("global", "all"):
    result["slash_command"] = uninstall_slash_command()
    if result["slash_command"].get("error"):
        result["errors"].append(result["slash_command"]["error"])   # <- flag not set
...
if purge and scope in ("global", "all") and global_cleanup_failed:
    result["purge_skipped"] = True
```

Reproduced with a genuine installed slash command (`_slash_is_ours` → `True`) and a read-only `~/.claude/commands`, so the `.md.bak` backup fails:

```
slash error   : Could not back up slash command: [Errno 13] Permission denied
slash removed : False        skipped_foreign: False
purge_skipped : False        purged: 2        errors: 1
slash file still present : True
~/.cozempic DELETED : True   ~/.cozempic_savings.json DELETED : True
```

The savings ledger and receipts are destroyed irreversibly while the install is still partially wired. The command does exit non-zero, but only after the deletion. `preview_uninstall` has the same gap — it collects `slash_error` separately and does not feed it into `purge_skipped` — so the dry run also promises the purge.

**The class is wider than this instance.** `global_cleanup_failed` is derived only from per-settings-file hook results, so every other global cleanup step can fail without blocking an irreversible purge:

| failing step | error recorded | feeds `global_cleanup_failed` | blocks purge |
|---|---|---|---|
| hook removal from a global settings file | yes | yes | yes |
| slash-command read / backup / removal | yes | **no** | **no** |
| global opt-out marker write | yes | **no** | **no** |
| remind-counter removal | yes | **no** | **no** |

The remind counter is genuinely cosmetic and reasonable to exclude, but the slash command and the opt-out marker are not — failing to write the opt-out marker means auto-init can re-wire, which is precisely the outcome the marker exists to prevent, and purging data in that state is the worst combination.

Suggested fix, which closes the whole class rather than the one instance: compute a single "global cleanup fully succeeded" value from every global-scope step, and have `preview_uninstall` and `run_uninstall` derive `purge_skipped` from that same value. A regression per row above would keep it closed. The rendering is already correct — `preview` prints "skipped due to a hook/settings error above" and the live path prints the matching message; only the predicate feeding them is under-inclusive.

## Medium — dry run and live disagree about purge when a settings file cannot be written

The two paths compute cleanup failure from different sources. `preview_uninstall` uses `cozempic_hook_schema_state(p).error`, which only *reads*; `run_uninstall` uses `uninstall_hooks(...)["error"]`, which reads *and writes*. A settings file that parses cleanly but cannot be rewritten is therefore invisible to the preview:

```
# genuine managed hook installed, ~/.claude made read-only so the backup fails
schema_state : found=True  error=None
DRY RUN      : purge_skipped=False   purge_data=2
LIVE         : purge_skipped=True    purged=0
               error: could not back up .../settings.json: [Errno 13] Permission denied
```

The live behaviour here is correct and fail-safe; the dry run is the wrong half. The direction is benign — live's predicate is a superset of the preview's, so the preview can only under-report skipping, never cause an unexpected deletion — which is why this is Medium and not part of the Critical above. It still defeats the purpose of a dry run for an irreversible operation: the user is told the ledger will be removed, runs it, and it is not.

Same categorical fix applies: one shared predicate used by both paths. Note this is reachable with no slash command involved at all, so fixing the Critical alone does not close it.

## The other three guarantees hold

Verified by execution, not inspection:

- **A managed global uninstall survives a failing local settings file.** With `settings.local.json` corrupted so its cleanup errors and `settings.json` clean: managed hooks removed, the error recorded, and `opt_out_set=True` with the marker present. The narrower `managed_global_cleanup_failed` gate does exactly what it claims. Worth the short comment noted elsewhere — the reason the marker gate is narrower than `global_cleanup_failed` is not obvious from the code.
- **Exit statuses.** Rejected purge confirmation → `SystemExit(1)`; any recorded error → `SystemExit(1)`; dry run with hook or slash errors → `SystemExit(1)`; `--purge --project` → `SystemExit(2)`.
- **Project-only dry runs stay project-scoped.** Slash command, remind counter, and purge output are all gated on `scope in ("global", "all")`, and `--purge --project` is rejected before the dry-run block. Confirmed for `scope="project"`: `purge_skipped=False`, no global rows emitted.

One fixture note that cost me two false starts and may be worth a line in the tests: a hand-written `settings.json` hook entry and a hand-written `cozempic.md` are both correctly classified as *not ours*, so a probe built that way silently exercises the no-op path (`removed: []` / `skipped_foreign: True`) and looks like a bug that is not there. Both fixtures above were therefore built with `wire_hooks` and `install_slash_command`.

## Verification (at `7924c74c`, in a detached worktree)

```
env -u COZEMPIC_NO_AUTO_INIT -u COZEMPIC_NO_AUTO_UPDATE uv run --frozen pytest -q --rootdir <checkout> \
  tests/test_uninstall.py tests/test_global_init.py tests/test_auto_init_global_skip.py \
  tests/test_doctor.py tests/test_hook_path_baking.py tests/test_nudge.py
-> 177 passed, 2 subtests passed
git diff --check -> clean
```

For the record, the seven `test_auto_init_global_skip.py` assertion failures reported earlier at this same head do not reproduce here — that file passes as part of the 177 above, run from inside the reviewed checkout with `--rootdir` pinned to it and `HOME` untouched. That looks like an invocation that let the ambient Claude directory leak into the fixture rather than a defect in the change.

All probes ran against a temporary `HOME`; no repository or user state outside the throwaway directories was modified.

eggrollofchaos (commented):
Reviewed at `fc1ea37` (current head). Summary: 1 Critical, 0 High, 0 Medium, 1 Low.

**The previously reported purge issue is fixed**

`run_uninstall` now sets `global_cleanup_failed` when slash-command cleanup errors, and `preview_uninstall` sets it on the corresponding `OSError`, so dry run and live agree for that failure mode. The hook path was already covered via `global_cleanup_failed or directory != "."`. Suite at this head: 77 passed, 2 subtests.

**Critical — the same data loss is still reachable through the opt-out write**

The guard was added at the slash-command branch, but the global opt-out write a few lines below records its failure without setting the flag:

```python
except OSError as exc:
    result["errors"].append(f"Could not persist global uninstall opt-out: {exc}")
    # global_cleanup_failed is not set here
```

So when `~/.claude` is not writable, `--purge` still proceeds. Reproduced in an isolated temp `HOME` with the hook and slash steps stubbed to succeed, so the opt-out write is the only failure:

```
errors        : ["Could not persist global uninstall opt-out: [Errno 13] Permission denied: .../ro/nested"]
opt_out_set   : False        <- auto-init can re-wire
purge_skipped : False
~/.cozempic still exists    : False
savings ledger still exists : False
```

This is the same outcome as the finding this PR set out to fix — `~/.cozempic` and the savings ledger deleted after an incomplete global uninstall — reached by a different trigger. It is arguably worse in effect: because the opt-out marker never persisted, auto-init can re-wire afterwards, so the user loses their data *and* the tool reinstalls. A nonzero exit does not mitigate it, since the deletion has already happened by then.

Suggested fix: set `global_cleanup_failed = True` in that `except`, mirroring the slash-command branch, and make the same change in `preview_uninstall` so the dry run keeps agreeing with the live path. Worth a regression test that stubs hooks and slash to succeed and fails only the opt-out write, so this specific combination is pinned.

**Low — the review request pointed at a superseded commit**

The request named `7924c74`, which is the head the earlier round already reviewed; the current head is `fc1ea37`, four "Address PR 177 review" commits later. I reviewed and bound the findings above to `fc1ea37`. Worth regenerating the target from the live head so a reviewer does not re-clear an old commit.

Not approving — leaving that to the PR owner.

eggrollofchaos (commented):
High — `cozempic init --uninstall-global` is documented as the deprecated alias for global uninstall, but it directly calls `uninstall_hooks()` only for `settings.json` and then sets the global opt-out marker. The current global-uninstall path now treats `settings.local.json` as a global hook file, so hooks there remain live after the alias reports success. Reproduced with managed hooks in both files: primary hook removed, local hook remains. Route the alias through `run_uninstall("global")` (or remove both global paths) and add coverage for this compatibility path.

eggrollofchaos (commented):
High — `src/cozempic/cli.py:1304`: the supported deprecated `init --uninstall-global` path removes only the configured `settings.json`, but global uninstall otherwise treats `settings.local.json` as managed too. A managed hook in that file survives while the global opt-out marker is written, so the command reports an uninstall even though protection remains active. Delegate the alias to `run_uninstall("global")` or clean both managed settings files before marking opt-out.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: not-applicable
- Cleared target: not-applicable
- Counts: Critical 0, High 1, Medium 0, Low 0, Nit 0
- Verdict: High finding remains

eggrollofchaos (commented):
Reviewed at `919fe62`. **1 Critical, 0 High, 0 Medium, 1 Low.**

**The previously reported purge issue is fixed and pinned.** Re-ran the earlier reproduction against this head, with hook and slash cleanup stubbed to succeed so the opt-out write is the only failure:

```
errors        : ["Could not persist global uninstall opt-out: [Errno 13] Permission denied: .../.cozempic_global_initialized"]
opt_out_set   : False
purge_skipped : True         <- was False
~/.cozempic still exists    : True    <- was deleted
savings ledger still exists : True    <- was deleted
```

Exact inversion of the earlier result. Reverting the one-line `global_cleanup_failed = True` fails `test_purge_is_skipped_when_global_opt_out_cannot_be_persisted`, so the behavior is genuinely test-pinned rather than incidentally passing. The deprecated-alias delegation is also covered — stubbing it out fails `test_deprecated_global_uninstall_removes_local_settings_too`. `cmd_uninstall` reads `all`, `project`, `purge`, and `dry_run` exclusively through `getattr` with defaults, so the hand-built `Namespace` cannot raise `AttributeError`, and routing through it means the alias now exits non-zero on error where the old path exited 0. 182 passed, 2 subtests at this head.

**Critical — the same data loss is still reachable through `--all`**

Enumerating every failure path in `run_uninstall` against the purge gate, one is still ungated. A project settings failure sets `project_cleanup_failed` but never `global_cleanup_failed`:

```python
if hook_result.get("error"):
    result["errors"].append(hook_result["error"])
    global_cleanup_failed = global_cleanup_failed or directory != "."   # project excluded
    project_cleanup_failed = project_cleanup_failed or directory == "."
```

`project_cleanup_failed` gates only the project opt-out marker; nothing downstream consults it. So under `uninstall --all --purge`, a project settings failure lets the purge run. Reproduced with the global targets succeeding and the project target failing:

```
errors        : ['simulated project settings write failure', ...]
opt_out_set   : True
purge_skipped : False
purged        : ['~/.cozempic', '~/.cozempic_savings.json']
~/.cozempic exists    : False
savings ledger exists : False
```

The end state is the failure mode this PR set out to close: hooks still wired in that project, `project_opt_out_set` false because the marker write is skipped when project cleanup failed, so project auto-init can re-fire — and all cozempic data deleted. `opt_out_set: True` only suppresses *global* auto-init, so the tool comes back and the savings ledger does not. Classifying this at the same severity as the earlier finding because the trigger differs but the outcome is identical. It is reachable through a documented flag combination and an ordinary state: `uninstall_hooks` returns an `error` dict for a malformed project `settings.json` rather than raising, which is exactly the case a user would reach for `--all --purge` to clean up.

The exit code does not mitigate it — deletion happens before `SystemExit(1)`, so the surfaced error tells the user what they already lost.

Suggested fix: gate the purge on any cleanup failure in scope, e.g. `if purge and scope in ("global", "all") and (global_cleanup_failed or project_cleanup_failed)`. Since `project_cleanup_failed` can only be true when a project target was in scope, this leaves `--global --purge` behavior unchanged. A regression test failing only the project target under `scope="all"` would pin it.

**Low — dry run still cannot predict the opt-out skip**

`preview_uninstall` tracks only `global_cleanup_failed` and has no opt-out write to fail, so `--dry-run --purge` reports that purge will proceed in exactly the case where the live run now skips it. Noting it as acknowledged and deferred rather than as a new problem: the divergence errs safe (the live run is the more conservative of the two), and for the `--all` gap above the two paths agree, since `preview_uninstall` mirrors the same project exclusion. Worth revisiting whenever the dry-run predictor is next touched.

Not approving — leaving that to the PR owner.

eggrollofchaos:
Addendum to the review above — one more finding, **Medium**. Revised counts: 1 Critical, 0 High, 1 Medium, 1 Low.

**`tests/test_uninstall.py` uninstalls cozempic from the checkout it runs in.**

`test_project_scope_purge_keeps_global_data` patches `HOME` but not the working directory:

```python
def test_project_scope_purge_keeps_global_data(self):
    (self.home / ".cozempic").mkdir()
    (self.home / ".cozempic_savings.json").write_text("{}")
    cz_init.run_uninstall("project", purge=True)
```

`run_uninstall("project", ...)` resolves project targets from `"."`, so the real checkout is the target. From a verified-clean tree, running that single test leaves:

```
 M .claude/settings.json          <- cozempic hooks stripped
?? .claude/.cozempic_uninstalled  <- project opt-out marker written
```

Bisected the file test-by-test; that is the only one that does it. Two consequences: a contributor running the suite silently loses cozempic in that checkout and gets the auto-init opt-out marker, and since `.claude/settings.json` is tracked, the stripped file sits staged-adjacent in `git status` where it can be committed by accident. CI is unaffected because the container is disposable, which is what keeps this from showing up in a green run.

Fix is a `chdir` into a temp directory for the project-scope case, matching how `HOME` is already isolated, or passing an explicit project directory rather than relying on the process cwd. Worth checking the other project-scope cases in the same file for the same assumption while you are in there.

Flagging it as Medium rather than higher because it damages a development environment, not user data or product behavior — but it is in scope here, since this PR adds the file.

eggrollofchaos (commented):
# Tier 3 Review — PR #177 (global hook migration)

Review tier: tier-3
Required review tiers: tier-1,tier-2,tier-3
Approval authority: requester
Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
Cleared target: 919fe6295254d5625ea144ec86eea044f7770c38
Counts: Critical 0, High 0, Medium 0, Low 0, Nit 0
Verdict: LGTM
Approval workflow: not-authority

## Review scope

Reviewed global and project hook-state handling, managed-versus-local settings behavior, uninstall/purge failure gates, opt-out marker persistence, the deprecated uninstall alias, doctor checks, and their regression coverage at the cleared target.

## Verification

- `git diff --check fd41e95b85f283286eaca9419c19825abb6d053b 919fe6295254d5625ea144ec86eea044f7770c38` — passed.
- Packet-recorded targeted suite: 182 passed, 2 subtests passed.

eggrollofchaos (commented):
## Review result

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 41457c8f3706215bc272f8ac03709ef9ee4b3bef
- Counts: Critical 0 | High 0 | Medium 0 | Low 0 | Nit 0
- Verdict: Clean — no Critical or High findings.
- Approval workflow: not-authority

Verified the uninstall scope split, fail-closed purge behavior, opt-out markers,
malformed-settings handling, and backup collision behavior. The targeted suite
passed: 186 tests plus 2 subtests; the CLI help smoke test also passed.

eggrollofchaos (commented):
## Review findings

No findings.

- Review tier: tier-3
- Required review tiers: tier-1,tier-2,tier-3
- Cleared base: fd41e95b85f283286eaca9419c19825abb6d053b
- Cleared target: 41457c8f3706215bc272f8ac03709ef9ee4b3bef
- Counts: Critical 0 | High 0 | Medium 0 | Low 0 | Nit 0
- Verdict: LGTM - approve
- Final-head clearance: tier-3 clean; requester authority must verify all selected tiers.
- Approval workflow: not-authority

Verification: the targeted uninstall/global-init/doctor/hook-path/nudge suite passed (186 tests, 2 subtests), `git diff --check` and `git fsck` passed, and current SonarCloud analysis is successful. PR context checked: head `41457c8f3706215bc272f8ac03709ef9ee4b3bef`; top-level comments 2; inline comments 0; review threads 0 (0 active); latest comment at 2026-07-27T10:07:51Z.

eggrollofchaos:
Correction to the submitted Tier 3 review: the submitted review, response, and response-ready signal remain preserved as historical evidence, but this correction makes the review non-clearing.

Review tier: tier-3
Required review tiers: tier-1,tier-2,tier-3
Approval authority: requester
Cleared base: not-applicable
Cleared target: not-applicable
Verdict: High findings remain.
Approval workflow: not-authority

High — `tests/test_uninstall.py:348-353` calls `run_uninstall("project", purge=True)` without switching to an isolated temporary project or mocking the destructive operation. Project-scope uninstall acts on the caller's current directory. Running the packet's prescribed test suite reproduced removal of the checkout's Cozempic hook entries from `.claude/settings.json` and created `.claude/.cozempic_uninstalled`. Run the test in a temporary project directory or mock the uninstall call; until then, normal local test execution can silently disable the protection under test.

eggrollofchaos (commented):
Reviewed at `41457c8f`. The purge guard reported earlier is fixed, and I re-verified it by execution rather than by reading the diff. One High remains — it confirms the isolation issue already raised against this head, but narrows it to a single line.

**1 High, 0 Critical, 0 Medium, 1 Low, 1 Nit.**

## The earlier purge finding is closed, verified per row

Rebuilt the fixtures with the real installers (`wire_hooks`, `install_slash_command`) in a throwaway `HOME`, then failed each global-scope cleanup step in turn:

| failing step | error recorded | `purge_skipped` | `~/.cozempic` + ledger survive |
|---|---|---|---|
| hook backup / settings write | yes | **true** | **yes** |
| slash-command backup | yes | **true** | **yes** |
| remind-counter removal | yes | **true** | **yes** |
| opt-out marker write | yes | **true** | **yes** |
| none (control) | no | false | purged, as intended |

All four rows now feed the guard (`init.py:890`, `908`, `918`, `929`) and the purge condition at `init.py:931` reads both the global and project flags. The control row confirms the guard is not simply blocking everything.

The dry-run parity claim also holds where it is scoped to hold. With a settings file that parses as a file but not as JSON — a failure the preview can detect by reading — preview and live agree: both report `purge_skipped=True`, and `purge_data` drops to 0. Same-second backup distinctness is correct too: five backups written inside one second produced five distinct names, all present, each retaining its own contents.

## High — `tests/test_uninstall.py:348` uninstalls cozempic from the directory the suite runs in

`test_project_scope_purge_keeps_global_data` patches `HOME` but never changes the working directory:

```python
def test_project_scope_purge_keeps_global_data(self):
    (self.home / ".cozempic").mkdir()
    (self.home / ".cozempic_savings.json").write_text("{}")
    cz_init.run_uninstall("project", purge=True)
```

`run_uninstall("project", …)` resolves its project targets from `"."`, so the target is the real checkout. Running the verification command from this PR's description in a clean checkout of this head mutated that checkout:

```
 M .claude/settings.json                       # cozempic hooks stripped
?? .claude/.cozempic_uninstalled               # opt-out marker planted
   .claude/settings.json.20260727_143535.bak   # backup of the pre-strip file
```

That marker is the auto-init opt-out, so the effect outlives the test run: cozempic stays unwired in that directory until someone notices and re-inits. It is silent — the suite reports `186 passed` either way — and it is reached by following the documented verification steps, which is how I hit it.

What makes this cheap to close is that the intended pattern is already used everywhere else. Every other project-scope call site in this file wraps the call in `os.chdir(project)` with a `finally: os.chdir(previous_cwd)`:

| call site | scope | cwd isolated |
|---|---|---|
| `:75` | `project` | yes (`:73`/`:77`) |
| `:100` | `project` | yes (`:98`/`:102`) |
| `:259` | `all`, purge | yes (`:257`/`:262`) |
| `:290` | `all`, purge | yes (`:285`/`:292`) |
| **`:351`** | **`project`, purge** | **no** |
| `:597` | `project` | yes (`:595`/`:599`) |

So the class is finite, fully enumerated, and one edit closes it: give `:351` the same temp-project `chdir` wrapper as its five siblings. Worth keeping the `purge=True` argument when you do — it is currently harmless only because project scope ignores purge, and the isolated version is what makes that safe to rely on.

## Low — the remind counter is the one path bound to the home directory at import time

`init.py:732` captures `_REMIND_COUNTER = Path.home() / ".cozempic_remind_counter"` at import. Every sibling path resolves at call time: the slash command, the settings paths, the opt-out marker, and the purge targets at `:935`. If the home directory changes between import and the call, uninstall acts on the wrong file and reports success:

```
bound at import      : HOME_A/.cozempic_remind_counter
home at call time    : HOME_B
counter in HOME_A    : deleted
counter in HOME_B    : still present
remind_counter_removed: True      errors: []
```

This is pre-existing — the constant is unchanged from the base commit, and this PR only adds the `try/except` around the unlink. I am reporting it as Low rather than higher because I could not name a caller that changes the home directory mid-process; in the normal CLI and hook paths it is fixed for the process lifetime.

Two things still make it worth a one-line fix here. This PR makes the purge guard depend on that row, so a wrong-home resolution means the row can never fail and never blocks a purge while a real counter survives. And the tests have to patch the constant (`tests/test_uninstall.py:35`) to work around it, which means the shipped resolution path is the one path the suite does not exercise. Turning it into a function, like `_global_init_marker()` and `_global_slash_path()` already are, removes both.

For the record, the same import-time capture exists at `digest.py:32`, `config.py:37`, and `updater.py:22-23` — all outside this diff. `digest.py:32` is worth a look on its own schedule, since `DIGEST_DIR` is the directory the purge deletes.

## Nit — the deferral note is narrower than the behaviour it defers

The description defers this as "dry-run cannot predict later write-only opt-out failures." The reasoning is right and I agree with the deferral, but the divergence is not limited to the opt-out marker: a hook-backup failure and a slash-command backup failure produce the same split (preview reports two items to purge, live skips). Worth widening the wording to write failures generally, so the next reader checking deferrals against behaviour does not find a case the note appears not to cover.

## Also verified

- The control row purges correctly, so the guard is not over-blocking.
- `186 passed, 2 subtests passed` reproduces at this head with `--rootdir` pinned to the checkout.
- `git diff --check` on `fd41e95b..41457c8f` is clean.
- Checks are green (`SonarCloud Code Analysis` pass); no inline comments and no active review threads.

All probes ran against throwaway home directories and a detached worktree; nothing outside them was modified, and the worktree that the suite dirtied was discarded.

eggrollofchaos (commented):
Reviewed the current lifecycle migration diff at this commit. No new blocking implementation findings. Focused lifecycle tests pass (186 collected); the previously reported project-scope uninstall test now runs in an isolated temporary directory.

PR #180, opened 2026-07-27
feat(digest): add COZEMPIC_NO_DIGEST opt-out

thematrixdev:
## Why

The behavioral digest is the only cozempic subsystem without an opt-out —
receipts have `COZEMPIC_NO_RECEIPTS`, telemetry has `COZEMPIC_NO_TELEMETRY`.
The plugin's hooks run `digest inject` on SessionStart/PostCompact and
`digest flush` on PreCompact/Stop unconditionally, so a user who doesn't want
extracted rules injected into Claude Code's memory has no supported way to turn
it off: deleting `cozempic_digest.md` just gets regenerated by the next hook run.

To be clear: we use and like cozempic — guard and treat/prune run on every
session here. This is about subsystem choice, for two concrete reasons:

1. Extraction quality: in our deployment the extractor occasionally
   misclassified ordinary instructions as behavioral corrections, sometimes
   inverting their meaning (e.g. "never mind, give me the command to oauth"
   became the standing rule "Do not ever mind, give me the command to oauth"),
   and those rules were then silently injected into every future session
   across projects.
2. Governance: some users manage Claude's behavioral rules through reviewed,
   version-controlled config (CLAUDE.md and similar) and want writes into
   Claude Code's memory directory to be an explicit choice rather than an
   always-on side effect of hooks.

An opt-out lets both kinds of users keep the parts of cozempic they want while
disabling the digest.

## What

- `COZEMPIC_NO_DIGEST` env var, mirroring `receipts_enabled` exactly: same
  truth table, same fail-closed handling of ambiguous non-empty values
  (privacy-style fail-safe: unrecognized opt-out intent disables).
- `digest_enabled()` gates every write path: `update_digest`, `flush_digest`,
  `recover_digest`, `sync_to_memdir`, and `save_digest_store` as the
  persistence choke point (covers store migration triggered by read paths).
- Opting out after rules were already synced removes the previously synced
  `cozempic_digest.md` and its generated MEMORY.md index line on the next
  hook-driven inject/recover — exact marker match only, so user-authored lines
  that merely mention the filename are preserved.
- `digest show` / `digest clear` keep working while disabled (inspection and
  cleanup of previously written state).
- CLI prints an explicit "disabled" message on mutating actions so hook logs
  show why nothing happened.
- README configuration row; tests covering the truth table, each gated path,
  the cleanup behavior, and an enabled-path control case; an autouse conftest
  fixture keeps the pre-existing digest tests hermetic on developer machines
  that export the variable.

No hooks.json change needed — the gate lives below the CLI, so the existing
hook commands become no-ops when the variable is set.


vdebellabre (commented):
Adopted this downstream (Windows 11 fork) — +1 on both the motivation and the mechanics. The extraction-quality failure mode described here matches what we observed verbatim (ordinary sentences inverted into standing prohibitions, injected cross-project from the global store). The opt-out works as documented: with `COZEMPIC_NO_DIGEST=1` baked into our hooks, the next hook-driven inject/flush removed the previously synced `cozempic_digest.md`, and the entry-shape-aware index cleanup correctly left user-authored `MEMORY.md` lines mentioning the file untouched. Matching the `COZEMPIC_NO_RECEIPTS` truth table (fail-safe direction included) is the right call.

Two Windows-only issues in the new test file, both one-line fixes we're running downstream:

1. `test_disabling_removes_previously_synced_memory` writes its `MEMORY.md` fixture with `Path.write_text(...)` and no `encoding=`. On Windows the default is the ANSI codepage (cp1252), so the fixture's em-dashes become `0x97` and `remove_synced_memory()`'s `read_text(encoding="utf-8")` raises `UnicodeDecodeError` — the test fails on Windows. Fix: `write_text(..., encoding="utf-8")` on both fixture writes.

2. The same test isolates home with `patch.dict(os.environ, {"HOME": home})`. On Windows `Path.home()` resolves `USERPROFILE`, not `HOME`, so the patch is inert — harmless in this test because `_get_memdir` is also patched, but the HOME-only pattern elsewhere in the suite is destructive (see the note on #182: `test_uninstall.py` deletes the developer's real global hooks on Windows). Suggest `{"HOME": home, "USERPROFILE": home, "COZEMPIC_NO_DIGEST": "1"}` so this new test starts life portable.

With those two tweaks the whole file passes on Windows (Python 3.14, Win11).

Issue #74, opened 2026-04-12
Possible sub-process spawn memory leak when pruning during guard ?


milankubin:
This is now the second time this has happened, and actually the second time pruning was needed (a /compact was done in between).
The first time I wasn't entirely sure if it was cozempic, but now the second time with the same pre-conditions  (a  session with a /compact and then subsequentially a prune) I can safely say it's a cozempic thread.

The pruning seems to have completed, but my user-slice gets evicted, everything gets killed off and lightdm2 restarts.
Which is normally typical for memory + swap being fully consumed.

Unfortunately, the sys journal is not being very helpful:

It starts with:
```
pr 12 11:29:33  systemd[3427]: Activating special unit exit.target...
apr 12 11:29:33  systemd[3427]: Stopping app-com.google.Chrome-3999.scope...
apr 12 11:29:33  systemd[3427]: Stopping app-org.chromium.Chromium-119519.scope...
apr 12 11:29:33  systemd[3427]: Stopping app-org.chromium.Chromium-120282.scope...
apr 12 11:29:33  systemd[3427]: Stopped vte-spawn-154f6592-8ab9-49dc-8b29-d033addab4ae.scope - VTE child process 122909 launched by mate-terminal proc>
apr 12 11:29:33  systemd[3427]: vte-spawn-154f6592-8ab9-49dc-8b29-d033addab4ae.scope: Consumed 1.150s CPU time, 34.1M memory peak, 1.3M memory swap pe>
apr 12 11:29:33  systemd[3427]: Stopped vte-spawn-5a3b1aa0-d970-479a-b609-5720a55f4023.scope - VTE child process 183133 launched by mate-terminal proc>
apr 12 11:29:33  systemd[3427]: vte-spawn-5a3b1aa0-d970-479a-b609-5720a55f4023.scope: Consumed 22min 17.762s CPU time, 2.7G memory peak, 571.8M memory>
apr 12 11:29:33  systemd[3427]: Stopped vte-spawn-89a01fd1-3333-49ac-b085-9356791b10b3.scope - VTE child process 140247 launched by mate-terminal proc>
apr 12 11:29:33  systemd[3427]: vte-spawn-89a01fd1-3333-49ac-b085-9356791b10b3.scope: Consumed 6.716s CPU time, 242.5M memory peak, 3.5M memory swap p>
[...]

PR #185, opened 2026-08-25
fix(guard): thread Claude PID through SessionStart; bound incremental scan reads

eggrollofchaos:
## Incident

Per-session process accumulation leak observed on macOS with Cozempic 1.8.39:

- **117-119 PPID-1 `cozempic.cli guard` processes survived concurrently.** The SessionStart hook backgrounds its entire updater/spawn subshell, so `cozempic guard --daemon` usually runs after reparenting under PID 1. `find_claude_pid()` then returns `None`, the daemon argv lacks `--claude-pid`, and the guard only checks owner exit when `claude_pid` is truthy — so these guards live forever despite the documented `Guard stopping (Claude exited)` path.
- **Hundreds of MiB to ~1 GiB RSS during 30-second polling** on 100-180 MiB transcripts. `load_messages_incremental()` is count-bounded (5000 messages) but not byte-bounded and reads the complete unread byte range into one `raw_bytes` allocation before parsing.

Upstream #74 fixed the mutation/prune path's peak memory; it did not touch this read-only polling path or the missing-owner lifecycle.

## Fix

**1. Guard owner lifecycle.** The hook now captures the identity-verified Claude PID (a process whose `comm` matches `claude`/`node` — the same check `find_claude_pid` uses) in its main shell, **before** the background boundary, and threads it to both the fresh-spawn and `--reload-self` guard paths via the existing hidden `--claude-pid` flag. `reload_self_daemon` accepts and forwards it. Fail-open unchanged: no claude ancestor → no `--claude-pid` → the guard falls back to its own detection. Hook schema bumped `v14 → v15` so installed hooks refresh.

**2. Bounded read-only checkpoint scan.** `load_messages_incremental()` is now streaming and byte-bounded (`_MAX_INCREMENTAL_READ_BYTES` per call). Complete lines are parsed, the unterminated tail deferred, and the byte offset resumed on the next poll. A single oversized JSONL line is skipped whole (consuming its line index), never buffered unboundedly. Partial-line deferral, surrogateescape byte accounting, rewrite/truncation invalidation, LRU, and the newest-N retention contract are all preserved.

## Tests

- Mutation-proof hook PID threading: a fake `claude` ancestor process (compiled C, so its `comm` contains "claude") drives the real hook through a backgrounded spawn and asserts the daemon argv carries `--claude-pid <the fake pid>`; deleting/bypassing the capture fails the test.
- Mutation-proof owner exit: `start_guard` with a live owner PID must exit when that PID dies on the first poll; deleting the watchdog block hangs the test.
- Bounded scan: sparse >200 MiB transcript proves the first read stays within budget; >10 MB single and partial trailing lines never defeat the bound; bounded convergence equals the newest-N of a full read.
- Full suite: 1899 passed. The 7 remaining failures are pre-existing/flaky (4 auto-init conftest interaction, 1 tmux-environment, 2 timing-sensitive stress tests) and pass in isolation on the pre-change base.

## Docs

`docs/guard-owner-census-proposal.md` — a read-only census/classification proposal for the pre-fix orphans, keyed by pidfile session UUID + recorded owner PID. No killing is performed by this change.

- Required review tiers: tier-1,tier-2

### DIFF
--- a/docs/guard-owner-census-proposal.md
+++ b/docs/guard-owner-census-proposal.md
@@ -1,0 +1,85 @@
+# Guard Owner-Census / Cleanup Proposal (read-only)
+
+**Status: proposal — read-only. This document and the commands it describes do
+NOT kill any process.** It describes how to enumerate and classify existing
+`cozempic.cli guard` daemons left over by the pre-fix SessionStart hook
+(ISSUE-AGT-637), so a FUTURE cleanup lane can act with confidence.
+
+## Background
+
+Before this fix, the SessionStart hook backgrounded its entire spawn subshell.
+`cozempic guard --daemon` therefore usually ran after reparenting under PID 1,
+`find_claude_pid()` returned `None`, and the daemon argv had no `--claude-pid`.
+Because the guard only checks owner exit when `claude_pid` is truthy, those
+guards never noticed Claude exiting and accumulated as PPID-1 orphans.
+
+Observed on macOS with Cozempic 1.8.39: 117-119 concurrent PPID-1
+`cozempic.cli guard` processes.
+
+This fix makes NEW guards correct (identity-verified owner PID threaded through
+the hook, guard exits when that PID dies). It deliberately does NOT add an
+auto-kill/reaper that guesses owners. Existing orphans are handled by a
+separate, human-gated cleanup lane driven by the census below.
+
+## Data available per guard
+
+| Source | Key | Holds |
+|---|---|---|
+| `/tmp/cozempic_guard_<slug>.pid` | `slug` = first 12 chars of session UUID | daemon PID (1-line, or 3-line newer format: pid on line 1) |
+| `/tmp/cozempic_guard_<slug>.log` | same `slug` | the spawn `CMD:` line, which carries `--claude-pid <pid>` for post-fix daemons (absent for pre-fix orphans) |
+| `/tmp/cozempic_guard_<slug>.startup-lock` | same `slug` | flock file (no payload) |
+
+There is **no persisted owner-PID record** on disk today: `_record_claude_identity`
+keeps `(pid, start_time)` only in the daemon's in-memory `_CLAUDE_IDENTITY` dict.
+The only durable owner evidence is the `--claude-pid` argument in the daemon's
+log `CMD:` line (post-fix only).
+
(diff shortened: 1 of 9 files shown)


eggrollofchaos (commented):
## Review findings

**High - `src/cozempic/session.py:1008`: a bounded cache miss returns stale head history as if it were current.** On a new cache entry for a transcript larger than `_MAX_INCREMENTAL_READ_BYTES`, the loader starts at byte 0, returns only the earliest chunk, and needs `ceil(size / 8 MiB)` guard polls to reach the current tail. `checkpoint_team()` immediately treats that partial list as authoritative, and the main loop derives `agents_active` from it. For a 100-180 MiB transcript, current team/subagent markers can therefore be invisible for about 6-12 minutes; if a hard threshold fires in that window, the guard can take a reload path while agents are actually active.

Please make first-read/rewrite semantics current-tail-safe: either load a bounded, newline-aligned tail so the first result represents the newest retained history, or explicitly mark an incomplete forward scan and make reload eligibility fail closed until the tail is reached. Add a mutation-sound regression with active-agent evidence only in the final chunk and assert the first guard decision cannot report inactive/reload-safe.

The owner-PID capture/threading path and bounded-read mechanics otherwise look sound. Verification: focused new tests `10 passed`; full suite `1906 passed, 5 skipped, 281 subtests passed`.

eggrollofchaos (commented):
## Follow-up review

Reviewed the repaired head `de5a104a0deba076e2ef035adcbb5cc5eb4b6cba` and the one-hop guard callers.

The prior High finding is resolved categorically: cache-miss, inode-replacement, truncation, and same-size rewrite paths now seed from a bounded, newline-aligned tail, so current team evidence is available on the first checkpoint instead of after a multi-poll forward scan. The incremental append path remains bounded, oversized lines retain their prior skip behavior, and callers that require complete history continue to use the non-incremental loader.

The focused regression covers active-agent evidence only at the tail and the first guard decision, plus rewrite/truncation and byte-boundary behavior. Reported verification: 13 focused tests and 38 adjacent atomic/leak tests passed; the four full-suite failures reproduce at the prior head and are unrelated.

Counts: Critical 0 | High 0 | Medium 0 | Low 0 | Nit 0

Verdict: clean at the current head.

